PR #33, opened 2022-05-27
Fix redis channel full

frostbyte73:


pkg/service/service.go
```diff
@@ -148,6 +150,11 @@ func (s *Service) handleRequest(req *livekit.StartEgressRequest) {
 		return
 	}
 
+	if s.handlingRoomComposite.Load() {
```
biglittlebigben, line 153: Could you explain why this is needed? It is not immediately obvious to me why the `sysload.CanAcceptRequest` test below is not sufficient.

frostbyte73: we only allow one room composite at a time, because it's hard to know how much CPU it will use and it can affect other recordings. It's probably safe to allow one room composite + some track or track composite, but leaving it like this for now until we're launching requests in separate processes and are happy with overall stability

biglittlebigben: Right, but I'm wondering if it would be more straightforward to add this check to the sysload package, possibly as part of the CanAcceptRequest call? Especially since we already somewhat track sessions with the AcceptRequest call?

pkg/service/service.go
```diff
@@ -242,6 +247,9 @@ func (s *Service) handleEgress(p *pipeline.Pipeline) {
 		case res := <-result:
 			// recording finished
 			s.sendEgressUpdate(res)
+			if s.handlingRoomComposite.Load() {
```
rrampage, line 250: Can also use AtomicBool's [compare and swap](https://pkg.go.dev/go.uber.org/atomic#Bool.CAS) here

pkg/service/service.go
frostbyte73: We could move it in there, but it would then become more of a request manager, we would need to tell it when the room composite completes, and would then also want to move `service.isIdle`, and `service.Status`. For now I think it's cleaner for that package to be CPU measurement only

biglittlebigben (approved):

PR #38, opened 2022-06-07
Add support for HLS segments and playlist output

biglittlebigben:
Add support for generating HLS segments and playlist as Egress output:
- This regenerates the m3u8 playlist for every new segment during the session. This helps with recovery in case of crash, and potentially allows publishing the HLS stream while live, as Event playlist.
- File upload and output info are not supported yet
- This uses `github.com/grafov/m3u8` as m3u8 implementation for now. This forces setting a maximum playlist size at creation, and can be inefficient when generating large playlists (the whole playlist is re-serialized each time)
- This uses the `mpegtsmux` gstreamer muxer. The muxer doesn't insert a PAT/PMT before each I-frame. This means that segments do not always start with a PAT/PMT. This can cause slower startup time on players that actually parse the PAT/PMT, such as VLC (the stream still works though). Most HLS players (including QuickTime) however seem to ignore the PAT/PMT, or search it in the whole segment at initialization
- This changes slightly the way we connect audio/video pads to the muxer module in the input part of the pipeline, iterating over several possible pad names until one matches. This is because the audio/video pad names are inconsistent across different gstreamer muxer plugins.


pkg/pipeline/params/vars.go
```diff
@@ -67,6 +71,7 @@ var (
 		OutputTypeMP4:  FileExtensionMP4,
 		OutputTypeTS:   FileExtensionTS,
 		OutputTypeWebM: FileExtensionWebM,
+		OutputTypeHLS:  FileExtensionMP4,
```
frostbyte73, line 74: Should this be a FileExtensionM3U8?

frostbyte73 (approved):
LGTM!

PR #39, opened 2022-06-08
Retry uploads

frostbyte73:


biglittlebigben (approved):
Thanks. Eventually, maybe we could use something such as https://github.com/flowchartsman/retry, with a context for cancellation, to handle better cases where the service is more heavily degraded and exponential backoff between retries may help.

biglittlebigben (approved):

PR #40, opened 2022-06-08
forward kill signal to spawned processes

frostbyte73:


biglittlebigben (approved):
LGTM. Is this actually needed? Unix systems tend to kill children processes when the parent is killed, unless extra care is taken when creating the Chile process. Not certain what the go runtime does...

frostbyte73:
> LGTM. Is this actually needed? Unix systems tend to kill children processes when the parent is killed, unless extra care is taken when creating the Chile process. Not certain what the go runtime does...

We catch the signal and send an EOS to the pipeline before exiting, so you don't lose what has already been recorded. The service waits for all subprocesses to complete before shutting down

PR #53, opened 2022-06-16
Cpu config and availability warnings

frostbyte73:


pkg/sysload/cpu.go
```diff
@@ -45,6 +51,63 @@ func Init(conf *config.Config, close chan struct{}, isAvailable func() float64)
...
+
+	recommendedMinimum := requirements[2]
+	if recommendedMinimum < 3 {
+		recommendedMinimum = 3
+	}
+
+	if numCPUs < requirements[0] {
+		recommended := requirements[2]
+		if recommended < 3 {
+			recommended = 3
```
davidzhao, line 92: can we reuse recommendedMinimum from above?

pkg/sysload/cpu.go
```diff
@@ -45,6 +51,63 @@ func Init(conf *config.Config, close chan struct{}, isAvailable func() float64)
...
+		}
+		logger.Errorw("not enough cpu", nil,
+			"minimum cpu", requirements[0],
+			"recommended", recommended,
+			"available", numCPUs,
+		)
+		return errors.New("not enough cpu")
+	}
+
+	if numCPUs < requirements[2] {
```
davidzhao, line 102: nit: this seems like an assumption that RoomComposite will always cost more.. it's a fair assumption, but I'm always wary about stuff like this.

it would also be better if we printed explicitly which requests this node will not be able to support. IMO it's more clear for someone trying to start a RoomComposite to see that it is explicitly not supported.

frostbyte73: they're sorted, it's not making that assumption

davidzhao: sorry, I missed that line

pkg/sysload/cpu.go
```diff
@@ -45,6 +51,62 @@ func Init(conf *config.Config, close chan struct{}, isAvailable func() float64)
...
+		logger.Errorw("not enough cpu", nil,
+			"minimum cpu", requirements[0],
+			"recommended", recommendedMinimum,
+			"available", numCPUs,
+		)
+		return errors.New("not enough cpu")
+	}
+
+	if numCPUs < requirements[2] {
+		logger.Errorw("not enough cpu for all request types", nil,
```
davidzhao, line 102:
```suggestion
		logger.Errorw("not enough cpu for some egress types", nil,
```

davidzhao (approved):

PR #57, opened 2022-06-22
Room composite startup speed

frostbyte73:


pkg/pipeline/params/params.go
```diff
@@ -95,12 +97,25 @@ type SegmentedFileParams struct {
...
-type UploadParams struct {
-	FileUpload interface{}
+func ValidateRequest(conf *config.Config, request *livekit.StartEgressRequest) (*livekit.EgressInfo, error) {
+	p, err := getPipelineParams(conf, request)
+	return p.Info, err
+}
+
+func GetPipelineParams(conf *config.Config, request *livekit.StartEgressRequest) (*Params, error) {
+	gstReady := make(chan struct{})
+	go func() {
```
biglittlebigben, line 107: I know that the current code structure doesn't expect the pipeline to be initialized more than once for each run of the executable, but to be on the forward looking side, it may make sense to use a sync.Once (or just a boolean since Params is not meant to be reentrant) to make sure we only ever run it once.

pkg/pipeline/params/params.go
```diff
@@ -109,6 +124,7 @@ func GetPipelineParams(conf *config.Config, request *livekit.StartEgressRequest)
 			RoomId:   request.RoomId,
 			Status:   livekit.EgressStatus_EGRESS_STARTING,
 		},
+		GstReady: make(chan struct{}),
```
biglittlebigben, line 127: Do we need this initialization both here and in GetPipelineParams? Could this lead to races where on waits on the wrong channel?

pkg/pipeline/params/params.go
frostbyte73: a new handler is launched for every request/pipeline, so it will always only be called once. I can move it to pipeline.New though, that only loses a few milliseconds

PR #84, opened 2022-08-18
Filenames

frostbyte73:
- Updated all the filename related params to Local_ or Storage_ to make it less confusing
- Remove leftover temp dirs after integration tests


biglittlebigben (approved):
Thsnks.

PR #93, opened 2022-08-30
Only use proxy pad with web source

frostbyte73:


biglittlebigben (approved):
Sure, but why is only the web source affected?

frostbyte73:
> Sure, but why is only the web source affected?

I honestly have no idea

PR #94, opened 2022-08-30
Fix segfaults

frostbyte73:
Proxy pad and buffers were occasionally being dereferenced and freed early, resulting in segmentation faults.
The issue is random but happened more frequently with track composite than room composite

### DIFF
--- a/pkg/pipeline/output/bin.go
+++ b/pkg/pipeline/output/bin.go
@@ -17,10 +17,11 @@ type Bin struct {
 	bin *gst.Bin
 
 	// stream
-	protocol params.OutputType
-	tee      *gst.Element
-	sinks    map[string]*streamSink
-	lock     sync.Mutex
+	protocol    params.OutputType
+	tee         *gst.Element
+	sinks       map[string]*streamSink
+	lock        sync.Mutex
+	isSDKSource bool
 
 	logger logger.Logger
 }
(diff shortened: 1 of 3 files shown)


pkg/pipeline/output/bin.go
```diff
@@ -80,32 +78,33 @@ func (b *Bin) Link() error {
...
-		if linkReturn := sink.queue.GetStaticPad("src").Link(proxy.Pad); linkReturn != gst.PadLinkOK {
-			return errors.ErrPadLinkFailed("rtmp sink", linkReturn.String())
+	sinkPad := sink.sink.GetStaticPad("sink")
+
+	// intercept FlowFlushing
+	proxy := gst.NewGhostPad("proxy", sinkPad)
+	proxy.Ref()
+
+	proxy.SetChainFunction(func(self *gst.Pad, _ *gst.Object, buffer *gst.Buffer) gst.FlowReturn {
+		buffer.Ref()
```
davidzhao, line 88: do we need to explicitly deref these anywhere?

davidzhao (approved):

pkg/pipeline/output/bin.go
frostbyte73: Go gst buffers are instantiated with a `runtime.SetFinalizer(wrapped, (*Buffer).Unref)` call before they're returned

PR #96, opened 2022-08-31
track vp8 webm instead of ivf

frostbyte73:


davidzhao:
mind updating README/docs in the same PR? 🙏

biglittlebigben (approved):

PR #99, opened 2022-09-07
Fix url building

frostbyte73:


pkg/pipeline/source/web.go
```diff
@@ -64,11 +64,20 @@ func NewWebSource(ctx context.Context, conf *config.Config, p *params.Params) (*
...
-	if err := s.launchChrome(ctx, inputUrl, p.Info.EgressId, p.Display, p.Width, p.Height, conf.Insecure); err != nil {
+	// build input url
+	inputUrl, err := url.Parse(p.TemplateBase)
+	if err != nil {
+		s.logger.Errorw("failed to parse template base", err)
+		s.Close()
+		return nil, err
+	}
+	values := inputUrl.Query()
+	values.Add("layout", p.Layout)
```
davidzhao, line 75: using `values.Set` would be better for these. If the user already passed those parameters in, we'd want to override them.

davidzhao (approved):

PR #112, opened 2022-09-16
more reasonable generated filenames

frostbyte73:
previously generated files as `room-2009-11-10 23:00:00 +0000 UTC m=+0.000000001.mp4`

### DIFF
--- a/pkg/pipeline/params/params.go
+++ b/pkg/pipeline/params/params.go
@@ -561,7 +561,7 @@ func (p *Params) updateFilepath(identifier string) error {
 
 	if p.StorageFilepath == "" || strings.HasSuffix(p.StorageFilepath, "/") {
 		// generate filepath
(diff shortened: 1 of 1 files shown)


pkg/pipeline/params/params.go
```diff
@@ -561,7 +561,7 @@ func (p *Params) updateFilepath(identifier string) error {
 
 	if p.StorageFilepath == "" || strings.HasSuffix(p.StorageFilepath, "/") {
 		// generate filepath
-		p.StorageFilepath = fmt.Sprintf("%s%s-%s%s", p.StorageFilepath, identifier, time.Now().String(), ext)
+		p.StorageFilepath = fmt.Sprintf("%s%s-%d%s", p.StorageFilepath, identifier, time.Now().Unix(), ext)
```
davidzhao, line 564: Unix() isn't very human friendly, can we do something like this?

`time.Now().Format("2006-01-02T150405")`

davidzhao (approved):

PR #115, opened 2022-09-21
Refactor, portrait presets

frostbyte73:
Refactored source and input for future participant composite egress
Added portrait presets


test/config.go
```diff
@@ -11,19 +11,20 @@ import (
...
+	"github.com/livekit/egress/pkg/config"
 	"github.com/livekit/egress/pkg/service"
 	"github.com/livekit/protocol/egress"
 	"github.com/livekit/protocol/utils"
 	lksdk "github.com/livekit/server-sdk-go"
-
-	"github.com/livekit/egress/pkg/config"
 )
 
-type Config struct {
```
biglittlebigben, line 22: `Context` usually has a fairly specific meaning in Go, so this was a bit confusing to me initially. Would TestPlan or TestSetup make sense here?

biglittlebigben (approved):
Small nit in a comment. We should also confirm that we are obj committing Participant Info related commented code as it announces we are working on it and may be harder to remove later if we change our mind.

frostbyte73:
> Small nit in a comment. We should also confirm that we are obj committing Participant Info related commented code as it announces we are working on it and may be harder to remove later if we change our mind.

Sure, I can always get the code back by finding this PR

PR #117, opened 2022-09-28
Update dockerfiles, fix handler signals

frostbyte73:
v1.4.1 appears to still be using the old gstreamer dockerfiles (builder had possibly cached them?)
Changed gstreamer dockerfile names to force update and can no longer reproduce the Caps changes not supported issue.

Also fixes handler shutdown, which was being killed instead of the signal being caught

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -36,7 +36,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.17
+          go-version: 1.18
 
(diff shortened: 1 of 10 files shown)


davidzhao:
Did we bump tag version after it was forked? if not, the cache theory would make sense.

davidzhao (approved):

frostbyte73:
> Did we bump tag version after it was forked? if not, the cache theory would make sense.

No, since the gst version was the same I just replaced livekit/gstreamer-1.20.3 on dockerhub

PR #128, opened 2022-10-11
handle participant disconnection

frostbyte73:


biglittlebigben (approved):

pkg/pipeline/input/sdk/appwriter.go
```diff
@@ -192,26 +192,29 @@ func (w *appWriter) start() {
...
-			// see comment in writeRTP below
-			if !w.clockSynced {
-				now := time.Now().UnixNano()
-				startTime := w.cs.GetOrSetStartTime(now)
-				w.ptsOffset = now - startTime
-				w.rtpOffset = int64(pkt.Timestamp)
-				w.clockSynced = true
-			}
+			// when a track is killed, one 8x8 frame is sent by the SFU and needs to be ignored
+			if pkt != nil {
```
davidzhao, line 196: this could ignore the first frame that comes through, is the 8x8 frame the first frame that comes down?

the concern is if we just dropped the first frame, and if it happened to be a keyframe (first real frame that client sends).. we could have just lost an RTT worth of data.

frostbyte73: not the first, it ignores the last

frostbyte73: once it reads a new packet into `next`, it pushes the previous `pkt`

davidzhao: oops.. brain fart on my end. you are right!

PR #134, opened 2022-10-13
Send eos to appwriter on failure

frostbyte73:
Fixes track recordings of screen shares (and is an alternative fix for track disconnection)

Will clean up queue/multiqueue related comments in a separate PR

### DIFF
--- a/pkg/pipeline/input/sdk/appwriter.go
+++ b/pkg/pipeline/input/sdk/appwriter.go
@@ -143,7 +143,8 @@ func (w *appWriter) start() {
 	// always post EOS if the writer started playing
 	defer func() {
 		if w.isPlaying() {
-			if flow := w.src.EndStream(); flow != gst.FlowOK {
+			flow := w.src.EndStream()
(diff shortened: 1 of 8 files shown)


davidzhao:
which issue does this resolve?

pkg/pipeline/input/sdk/appwriter.go
```diff
@@ -192,29 +192,26 @@ func (w *appWriter) start() {
 				return
 			}
 
-			// when a track is killed, one 8x8 frame is sent by the SFU and needs to be ignored
-			if pkt != nil {
```
davidzhao, line 196: we no longer need this?

frostbyte73: Nope. This was an attempt to get around the final packet and worked for killing a video track, but doesn't work for screen share. The screen share fix of allowing the error and sending an eos to that appwriter also works on regular video tracks

frostbyte73:
> which issue does this resolve?

Updated description

davidzhao (approved):
nice!

PR #141, opened 2022-10-18
Clean up active counter

frostbyte73:
not needed, since each appsrc handles its own stream and sends EOS events

### DIFF
--- a/pkg/pipeline/pipeline.go
+++ b/pkg/pipeline/pipeline.go
@@ -281,9 +281,19 @@ func (p *Pipeline) messageWatch(msg *gst.Message) bool {
 		}
 
 		p.Logger.Debugw("EOS received, stopping pipeline")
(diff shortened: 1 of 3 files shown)


pkg/pipeline/pipeline.go
```diff
@@ -281,9 +281,19 @@ func (p *Pipeline) messageWatch(msg *gst.Message) bool {
 		}
 
 		p.Logger.Debugw("EOS received, stopping pipeline")
-		if p.Info.Status == livekit.EgressStatus_EGRESS_ACTIVE {
-			p.Info.Status = livekit.EgressStatus_EGRESS_ENDING
-		}
+		p.closeOnce.Do(func() {
```
biglittlebigben, line 284: Can this block be refactored in a subroutine since we have basically the same code in SendEOS?

biglittlebigben (approved):

biglittlebigben (approved):

PR #146, opened 2022-10-21
Mix silence in with sdk audio

frostbyte73:


pkg/pipeline/input/builder/audio.go
```diff
@@ -0,0 +1,299 @@
...
+		return err
+	}
+	a.testSrc = []*gst.Element{audioTestSrc, audioCaps}
+
+	audioMixer, err := gst.NewElement("audiomixer")
+	if err != nil {
+		return err
+	}
+	// set latency slightly higher than max audio appsrc latency
+	if err = audioMixer.SetProperty("latency", uint64(41e8)); err != nil {
```
davidzhao, line 233: should these latency values be constants or configurable?

pkg/pipeline/input/builder/audio.go
```diff
@@ -0,0 +1,299 @@
...
+		return errors.ErrNotSupported(string(p.AudioCodec))
+	}
+
+	return nil
+}
+
+func getCapsFilter(p *params.Params) (*gst.Element, error) {
+	var caps *gst.Caps
+	switch p.AudioCodec {
+	case params.MimeTypeOpus:
```
davidzhao, line 278: should we also set websocket / raw output to two channel and 48k for consistency for real-time workflows?

davidzhao (approved):
👍

pkg/pipeline/input/builder/audio.go
frostbyte73: it's already doing that, audio codec is opus in that case with output type raw

Issue #142, opened 2022-10-18
track + trackComposite issues (writing audio).


neilkinnish:
I'm noticing an issue with `track` and `trackComposite`. 

I'll try and explain – I'm still investigating – (I have files which I can supply).

Example: Audio recording which is 32 seconds where the user spoke through out. Resulting file has the correct duration, but the audio is within the first half of the recording.

I can confirm that if I record the `roomComposite` (as well) or/and add video to the `trackComposite` then the audio playback is fine, although additionally I am seeing drift on occasion in this area.

My assumption right now is just how the frames are being written, but I haven't looked deeper at this stage. I wanted to raise this issue first to see if there is any knowledge of this before I look deeper.

neilkinnish:
To clarify, the audio is not sped up, more like silent frames are being dropped.

neilkinnish:
UPDATE: I've been looking a little deeper and it seems every audio only recording is affected.

frostbyte73:
A few questions, since I'm having trouble reproducing this:
Are the users muting themselves, or just not speaking sometimes? 
Which version of egress are you running? 
What player are you using?

neilkinnish:
Hi @frostbyte73 thanks for looking into this.

1. Muted: No, I have been testing on my own and with people just counting with pauses (no muting) I have some file outputs to demonstrate the issue if it helps.
2. Egress: We are currently on 1.4.5 but have also tested custom build from master.
3. Player: vlc, browser, etc. We are also checking/converting the files in ffmpeg.

Some additional info on some more recordings made today....

File reporting duration of 34 seconds, but actually playback is 24 seconds and different players will report either one (or in some cases both).

We have checked the output `ogg` file from the egress in `opusdec` and `opusinfo`.

opusdec is reporting quite a few `Warning: Hole in data.`

opusinfo is reporting:

```
WARNING: sequence number gap in stream 1. Got page 15 when expecting page 14. Indicates missing data.
WARNING: discontinuity in stream (1)
```

Hope this helps.

neilkinnish:
@frostbyte73 I’ve been running a few tests with the sample builder (actually I’m doing a quick test with pion sample builder against livekit) very quick and rough. First I was logging silence data and then I adjusted to fill and this is solving the issue I believe. I’m not familiar with the differences in livekits sample builder at a quick glance I can see it’s making different decisions.

frostbyte73:
yup, currently working on mixing silence in

frostbyte73:
[...]

PR #148, opened 2022-10-23
fix: silence frame

neilkinnish:
This addresses a new issue created since recent changes to fix silence data being added https://github.com/livekit/egress/issues/142.

The current implementation results in audio cut out following silence.

This PR fixes that issue (in my tests).


neilkinnish:
To add.... this needs to be tested, it seems to be solving my issue in tests. But I'm not convinced it's addressing the core issue.

neilkinnish:
Closing this, needs more investigation. But there is definitely an issue. I'll raise a new ticket.

PR #157, opened 2022-10-27
Update readme and filenames

frostbyte73:


pkg/pipeline/params/params.go
```diff
@@ -498,13 +500,24 @@ func (p *Params) updateFileParams(storageFilepath string, output interface{}) er
...
+		replacements = map[string]string{
+			"{room_name}": p.Info.RoomName,
+			"{room_id}":   p.Info.RoomId,
+			"{time}":      time.Now().Format("2006-01-02T150405"),
+		}
+	} else {
+		identifier = "web"
+		replacements = map[string]string{
+			"{time}": time.Now().Format("2006-01-02T150405"),
+		}
```
biglittlebigben, line 516: Is it possible to put this block in a subroutine since it is reused?

davidzhao (approved):

PR #186, opened 2022-12-12
Create copy of base config, redact upload keys/secrets

frostbyte73:


pkg/config/pipeline.go
```diff
@@ -153,25 +153,21 @@ func (p *PipelineConfig) Update(request *livekit.StartEgressRequest) error {
...
 
-	p.updateUploadConfig()
+	p.updateUploadConfig(request)
+	connectionInfoRequired := true
 
 	switch req := request.Request.(type) {
 	case *livekit.StartEgressRequest_RoomComposite:
 		p.Info.Request = &livekit.EgressInfo_RoomComposite{RoomComposite: req.RoomComposite}
 		p.Info.RoomName = req.RoomComposite.RoomName
-		if p.Info.RoomName == "" {
```
biglittlebigben, line 162: Is this check already made somewhere else or is it not needed?

biglittlebigben (approved):

pkg/config/pipeline.go
frostbyte73: moved further down

PR #188, opened 2022-12-13
Do not duplicate frames to reach the video profile target framerate. Drop frames if the encoder can't keep up

biglittlebigben:
This PR makes sure that the videorate plugin only drops frame if the input framerate is higher than the target, and does not duplicate frames to reach the target framerate in all cases.
This also marks the post decoder queue as leaky to avoid pipeline stalls if the encoder can't keep up.


biglittlebigben:
The updated tests may fail because the xvimagesrc can't keep up 30fps, not that we don't duplicate frames, but I'll need to make sure on a better network. I'll update the test if so.

biglittlebigben:
We should not publish this to cloud immediately and get some testing on it first.

frostbyte73 (approved):

PR #192, opened 2022-12-14
Egress debug HTTP handler 

biglittlebigben:
This PR adds an optional debug HTTP handler to egress. Currently, the handler only has one feature, which is taking snapshots of the gstreamer pipeline for a given egress in the dot format.


biglittlebigben:
This works by opening a http server listening on the localhost interface on a different port for each handler process. The main http server on the egress service proxies the call to the TCP port assigned to the handler for the requested egress.

biglittlebigben:
Using unix sockets for the communication between the service and the handler may be a cleaner option, and it looks like you can use the go http implementation over unix sockets (with mixed reports of success). Should we tackle this now or is this something addressable later?

pkg/service/handler.go
```diff
@@ -92,10 +102,47 @@ func (h *Handler) Run() error {
 			}
 
 			h.sendResponse(ctx, request, p.Info, err)
+		case debugResponseChan := <-h.debugRequests:
```
frostbyte73, line 105: can this happen inside Handler.GetPipelineDebugInfo instead?

pkg/config/service.go
```diff
@@ -33,6 +35,10 @@ type CPUCostConfig struct {
 	WebCpuCost            float64 `yaml:"web_cpu_cost"`
 }
 
+type DebugConfig struct {
+	DebugHandlerPort int `yaml:"debug_handler_port"` // Port used to launch the egress debug handler. 0 means disabled. The egress service will reserve the next 1000 port to exchange debug info with the egress handlers.
```
frostbyte73, line 39: could avoid extra nesting and put it with healthport/prometheus port

pkg/service/http.go
```diff
@@ -0,0 +1,86 @@
...
+	"strings"
+
+	"github.com/livekit/egress/pkg/errors"
+)
+
+const (
+	gstPipelineDotFile = "/gst_pipeline/"
+)
+
+type handerProxyHandler struct {
```
frostbyte73, line 16: typo?

frostbyte73:
we should consider using grpc over sockets instead of http, especially if we plan on passing any other info between service and handlers

pkg/config/service.go
biglittlebigben: Yes

pkg/service/http.go
biglittlebigben: Yes

biglittlebigben:
> we should consider using grpc over sockets instead of http, especially if we plan on passing any other info between service and handlers
> 
> I have it already implemented for other things on the `prelaunch` branch if we want to use that [c5fa328](https://github.com/livekit/egress/commit/c5fa328e192b2ccc1ea4cef9a883feccecda14a7)

That's more forward looking. I'll check if I can do this quickly now. If not that can easily be changed later as service and handler are always shipped together.

pkg/service/handler.go
biglittlebigben: Yes, if we store a reference to the pipeline inside the Handler object, protected by a mutex, with a null check in GetPipelineDebugInfo. Both options have pros and cons I think and I have no strong opinion. What do you prefer?

biglittlebigben: More specifically: the Mutex option is less code, and less indirections. The channel option may be a more Go like pattern, and I. think there is some readability gained by keeping all accesses to an object within the same function/runloop.

magefile.go
```diff
@@ -47,7 +106,7 @@ func Integration(configFile string) error {
 
 	return mageutil.Run(context.Background(),
 		"docker build -t egress-test -f build/test/Dockerfile .",
-		fmt.Sprintf("docker run --rm -e EGRESS_CONFIG_FILE=%s -v %s/test:/out egress-test", configFile, dir),
+		fmt.Sprintf("docker run --rm -e EGRESS_CONFIG_FILE=%s -p 9000:9000 -v %s/test:/out egress-test", configFile, dir),
```
frostbyte73, line 109: is the port still needed?

frostbyte73 (approved):

pkg/service/http.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+	default:
+		http.Error(w, "unkown application", http.StatusNotFound)
+		return
+	}
+
+	grpcResp, err := p.processManager.sendGrpcDebugRequest(egressId, grpcReq)
+	switch {
+	case errors.Is(err, errors.ErrEgressNotFound):
+		http.Error(w, "unkown egress", http.StatusNotFound)
```
frostbyte73, line 45: one more typo

pkg/service/service.go
```diff
@@ -103,6 +103,25 @@ func (s *Service) Run() error {
...
+
+	mux := http.NewServeMux()
+	mux.Handle(fmt.Sprintf("/%s/", gstPipelineDotFile), h)
+
+	go func() {
+		addr := fmt.Sprintf(":%d", s.conf.DebugHandlerPort)
+
+		logger.Debugw(fmt.Sprintf("starting debug handler on address %s", addr))
+		err := http.ListenAndServe(addr, mux)
+		logger.Infow("debug server failed", "error", err)
```
frostbyte73, line 121: ListenAndServe always returns a non-nil error, probably don't want to log it

biglittlebigben: From my testing it doesn't return when the process/service gets killed, so we should not see the log message unless the server fails to start. I'll replace it with an Info message though

Issue #196, opened 2022-12-16
Egress record start failed with pulse sink load module Failure: Module initialization failed


tomxiong:
Hi team,
I found that some time the egress cannot start new record with error Module initialization failed. After doing some investigation, I found that the egress task will not more than 60 due to it will cause a lot of suspended pulse module. 
The log error like below:
```
{"level":"debug","ts":1670910590.9727807,"logger":"egress","caller":"web/source.go:172","msg":"chrome log: START_RECORDING","nodeID":"NE_A5mg2fqY267D","egressID":"EG_GDv7NBjewXdG"}
{"level":"error","ts":1670910591.1060054,"logger":"egress","caller":"pipeline/pipeline.go:207","msg":"failed to set pipeline state","nodeID":"NE_A5mg2fqY267D","egressID":"EG_GDv7NBjewXdG","error":"Failed to change state to PLAYING","stacktrace":"github.com/livekit/egress/pkg/pipeline.(*Pipeline).Run\n\t/workspace/pkg/pipeline/pipeline.go:207\ngithub.com/livekit/egress/pkg/service.(*Handler).HandleRequest.func2\n\t/workspace/pkg/service/handler.go:58"}
{"level":"debug","ts":1670910591.1061406,"logger":"egress","caller":"pipeline/pipeline.go:699","msg":"removing temporary directory","nodeID":"NE_A5mg2fqY267D","egressID":"EG_GDv7NBjewXdG","path":"/out/data/EG_GDv7NBjewXdG/"}
{"level":"info","ts":1670910591.1068933,"logger":"egress","caller":"service/handler.go:125","msg":"egress failed","nodeID":"NE_A5mg2fqY267D","egressID":"EG_GDv7NBjewXdG","request_type":"room_composite","output_type":"file","error":"Failed to change state to PLAYING"}
{"level":"debug","ts":1670910591.1263373,"logger":"egress","caller":"service/service.go:270","msg":"deleting handler temporary directory","nodeID":"NE_nWkNtRgAdZUz","path":"/tmp/lkegress/EG_GDv7NBjewXdG"}
{"level":"debug","ts":1670910967.866186,"logger":"egress","caller":"service/service.go:168","msg":"request received","nodeID":"NE_nWkNtRgAdZUz","egressID":"EG_PH6EAR4ZSWDa","requestID":"RPC_uBpivEa8Nsg7","senderID":"ND_KU22jczb"}
[...]

PR #197, opened 2022-12-16
Egress record start failed with pulse sink load module Failure: Module initialization failed #196

tomxiong:
* Remove the \n  character from the return result
*  I add debug log after WebInput.close() to show the module NO. which it wants to unload-module and found that the "pulseSink" has value include "\n". And I try to execute the command "pactl unload-module 10\n"at egress server, it returns error:
```
root@qm-egress-deployment-bbcf5fbc6-7p5fq:/data# pactl unload-module 10\n
Failed to unload module: Module 10n not loaded
```
* So I think that the "pulseSink" is comes from the result of "pactl load-module ...", but it contains the \n character. So we can remove it before set it to "pulseSink". 
* I test the changed code in my environment, and it works well, and no more new modules left in pulse audio after egress stop.

@davidzhao , @frostbyte73 , please help to review the code.

### DIFF
--- a/pkg/pipeline/input/web/source.go
+++ b/pkg/pipeline/input/web/source.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/url"
 	"os/exec"
+	"strings"
 
 	"github.com/chromedp/cdproto/runtime"
 	"github.com/chromedp/chromedp"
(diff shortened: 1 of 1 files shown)


biglittlebigben (approved):
Thanks for the investigation and the patch. We will merge this.

PR #200, opened 2022-12-21
Do not transcode H264 video track egress

biglittlebigben:


pkg/pipeline/input/builder/video.go
```diff
@@ -107,7 +107,7 @@ func (v *VideoInput) buildSDKDecoder(p *config.PipelineConfig, src *app.Source,
 	case strings.EqualFold(codec.MimeType, string(types.MimeTypeH264)):
 		if err := src.Element.SetProperty("caps", gst.NewCapsFromString(
 			fmt.Sprintf(
-				"application/x-rtp,media=video,payload=%d,encoding-name=H264,clock-rate=%d",
+				"application/x-rtp,media=video,payload=%d,encoding-name=H264,clock-rate=%d,width=1980,height=1080",
```
frostbyte73, line 110: p.Width/Height?

frostbyte73 (approved):

pkg/pipeline/input/builder/video.go
davidzhao: FYI: it's possible for TrackInfo.width to be 0 sometimes. It's a rare condition. we should have some defaults for this if possible.

Also, do the values set here do anything? or is it just to get the downstream caps to be happy?

biglittlebigben: Thanks for catching this. This is test code that made its way in the PR. the h264parse step will extract the actual dimensions.

davidzhao (approved):
lgtm

biglittlebigben:
Track composite allows you to force transcoding. This will however not do framerate conversion unless the source framerate is above the encoder framerate setting. You'll need to use RoomComposite to reach a constant framerate. Framerate conversion will however inherently create a somewhat jittery output (which is why we try to avoid it when possible)

PR #201, opened 2022-12-23
Use separate queues for audio and video into the muxer. Fix integration test performance

biglittlebigben:
This PR switches the multi queue for 2 separate queues for audio and video. It also reduces the length of the input raw buffer queues to decreases memory usage.

In order to improve the performance of the integration test suite, run against an optimized version of Gstreamer.

### DIFF
--- a/build/test/Dockerfile
+++ b/build/test/Dockerfile
@@ -4,6 +4,33 @@ WORKDIR /workspace
 
 ARG TARGETPLATFORM
 
+# install go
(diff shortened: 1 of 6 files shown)


frostbyte73 (approved):

frostbyte73:
Tested with youtube. Still get backwards dts warnings, but the stream sounds good. Merging

biglittlebigben:
OK, thanks. I'm planning to have a look too later today.

PR #202, opened 2022-12-23
Use psrpc

frostbyte73:


cmd/server/main.go
```diff
@@ -83,13 +90,28 @@ func runService(c *cli.Context) error {
...
-	rc, err := redis.GetRedisClient(conf.Redis)
+	rc, err := lkredis.GetRedisClient(conf.Redis)
 	if err != nil {
 		return err
 	}
 
-	rpcServer := egress.NewRedisRPCServer(rc)
-	svc := service.NewService(conf, rpcServer)
+	ctx := context.Background()
+	previous, err := rc.Get(ctx, ev.EgressVersionKey).Result()
```
davidzhao, line 99: what happens when multiple egress instances are running? like as part of a rolling deploy?

pkg/service/service.go
```diff
@@ -103,104 +113,71 @@ func (s *Service) Run() error {
...
+func (s *Service) StartEgressAffinity(req *livekit.StartEgressRequest) float32 {
+	if !s.manager.canAccept(req) || !s.monitor.CanAcceptRequest(req) {
+		return 0
+	}
 
-		s.sendResponse(ctx, req, p.Info, err)
-		if err != nil {
-			span.RecordError(err)
-		}
+	if s.manager.isIdle() {
```
davidzhao, line 141: what do these constants mean? comments here would be helpful.

pkg/service/service.go
```diff
@@ -14,34 +14,44 @@ import (
...
+		monitor:     monitor,
+		manager:     NewProcessManager(conf, monitor),
+		shutdown:    make(chan struct{}),
 	}
+
+	psrpcServer, err := rpc.NewEgressInternalServer(conf.NodeID, s, bus)
+	if err != nil {
+		return nil, err
+	}
+	s.psrpcServer = psrpcServer
```
davidzhao, line 53: how is this used? I don't see references to psrpcServer anywhere.. does it just get started automatically?

test/integration.go
```diff
@@ -377,7 +404,13 @@ func runSegmentsTest(t *testing.T, conf *TestConfig, req *livekit.StartEgressReq
 
 func startEgress(t *testing.T, conf *TestConfig, req *livekit.StartEgressRequest) string {
 	// send start request
-	info, err := conf.rpcClient.SendRequest(context.Background(), req)
+	var info *livekit.EgressInfo
+	var err error
+	if conf.PSRPC {
+		info, err = conf.psrpcClient.StartEgress(context.Background(), req)
```
davidzhao, line 410: nice perk of having matching method names.

pkg/service/service.go
frostbyte73: added a shutdown method to drain and close

davidzhao: so it does start automatically when the server is created?

frostbyte73: yeah

davidzhao (approved):

cmd/server/main.go
frostbyte73: I think this should be ok with redis? I could add locking, but for that we just use a SetNX and check which one got the write

PR #208, opened 2022-12-30
Add pprof debug handler

biglittlebigben:
This allows running supported pprof profiling tools either on the main service, or any of the running handlers.

Also validate that we have a valid egress id in start requests


davidzhao (approved):
lg

frostbyte73 (approved):

PR #213, opened 2023-01-09
Use sender reports for track sync

frostbyte73:


pkg/pipeline/input/sdk/synchronizer.go
```diff
@@ -1,39 +1,197 @@
...
+					t.Unlock()
+				}
+			}
+		} else {
+			t := c.syncInfo[pkt.SSRC]
+			pts, _ := t.getPTS(int64(pkt.RTPTime))
+			expected := mediatransportutil.NtpTime(pkt.NTPTime).Time().Sub(c.ntpStart)
+			if pts != expected {
+				diff := expected - pts
+				if diff > time.Millisecond*20 {
```
biglittlebigben, line 124: Can it happen that pts > expected? If so, should we consider abs(diff)?

biglittlebigben (approved):

pkg/pipeline/input/sdk/synchronizer.go
frostbyte73: ah, yes. Happens frequently

PR #220, opened 2023-01-19
Ignore errors from splitmuxsink after EOS is sent

biglittlebigben:
This works around a GST race where splitmuxsink would send an error event while we are tearing the pipeline out.

### DIFF
--- a/pkg/pipeline/pipeline.go
+++ b/pkg/pipeline/pipeline.go
@@ -37,6 +37,7 @@ const (
 
 	elementGstRtmp2Sink = "GstRtmp2Sink"
 	elementGstAppSrc    = "GstAppSrc"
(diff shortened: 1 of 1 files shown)


frostbyte73 (approved):

frostbyte73:
Does this only happen when there's been no data? (Is there any situation where we'd ignore an actual failure to close out a playlist?

biglittlebigben:
> Does this only happen when there's been no data? (Is there any situation where we'd ignore an actual failure to close out a playlist?)

This tests for any kind of child mixer failure after EOS is sent indeed. I've only seen them in the case above, but I imagine it could happen for other errors. Impact would be the loss of the last segment only. We still log the error as Debug so investigation of a bug report should still be possible.

PR #223, opened 2023-01-20
sanity check pts drift

frostbyte73:


pkg/pipeline/input/sdk/synchronizer.go
```diff
@@ -116,37 +131,47 @@ func (c *synchronizer) onRTCP(packet rtcp.Packet) {
...
+func (t *syncInfo) onSenderReport(pkt *rtcp.SenderReport, ntpStart time.Time) {
+	pts, _ := t.getPTS(int64(pkt.RTPTime))
+	expected := mediatransportutil.NtpTime(pkt.NTPTime).Time().Sub(ntpStart)
+	if pts != expected {
+		diff := expected - pts
+		apply := true
+		t.Lock()
+		if absGreater(diff, largePTSDrift) {
+			logger.Warnw("high pts drift", nil, "trackID", t.trackID, "pts", pts, "diff", diff)
+			if absGreater(diff, massivePTSDrift) {
```
biglittlebigben, line 154: With this new algorithm, it looks like we'll still fail the export if there is a jump in the timestamp somewhere that lasts more than 1 RTCP report?

biglittlebigben (approved):

boks1971 (approved):
I think the problem needs to be better handled by SFU synchronising sender reports across all down tracks.

pkg/pipeline/input/sdk/synchronizer.go
frostbyte73: If they're consistent, then I think a large drift could be possible/valid in a situation where the video track begins muted and we don't know the framerate

PR #232, opened 2023-01-31
use s3 uploader

frostbyte73:
will automatically create multipart uploads for large files


biglittlebigben (approved):
That's a convenient object from S3... Do we need a README update about setting up unused part deletion policies on the S3 bucket?

frostbyte73:
> That's a convenient object from S3... Do we need a README update about setting up unused part deletion policies on the S3 bucket?

I don't think so, if it fails it'll call `AbortMultipartUpload`

PR #238, opened 2023-02-02
Have the template server take the template FS as argument

biglittlebigben:


frostbyte73:
can you open a cloud-egress PR as well pointing to this branch, so we can see if the build/test succeeds?

frostbyte73 (approved):

PR #248, opened 2023-02-08
Better url validation, redact stream keys

frostbyte73:


pkg/config/output.go
```diff
@@ -486,6 +485,29 @@ func redactEncodedOutputs(out interface {
...
+	rtmpRegexp = regexp.MustCompile("^(rtmps?:\\/\\/)(.*\\/)(.*\\/)(\\S*)( live=1)?$")
+)
+
+func redactStreamKey(url string) (string, bool) {
+	match := rtmpRegexp.FindStringSubmatch(url)
+	if len(match) != 6 {
+		return url, false
+	}
+
+	match[4] = redact(match[4])
```
biglittlebigben, line 507: On the ingress side, I've seen folks skipping the "application" part of the RTMP url, ie using 

`rtmp://<server>/<stream_key>`

OBS supports sending such content at least. I'm not sure if any service would support such a URL (beyond our ingress...), but it looks like we would not redact it?

biglittlebigben (approved):

pkg/config/output.go
frostbyte73: rtmp2sink requires it

frostbyte73: AFAICT the stream providers always include the application part as well (I checked twitch, youtube, and facebook). When I switched from rtmpsink to rtmp2sink, I was forced to change the local test streams from `rtmp://localhost:1935/stream1` to `rtmp://localhost:1935/live/stream1`

PR #258, opened 2023-02-13
Multiple outputs

frostbyte73:


biglittlebigben (approved):
Should we add a h264parse step to stream and file in case we ever add the outputs in a different order?

frostbyte73:
> Should we add a h264parse step to stream and file in case we ever add the outputs in a different order?

caps negotiation happens after everything is added, I think it should be ok

PR #264, opened 2023-02-22
Reset frozen timer when opening/closing segments

frostbyte73:


biglittlebigben (commented):
Accessing the `eosTimer` struct member from the message handler seems slightly iffy from a concurrency perspective, but I get we already do in the EOS message handler... The documentation for `timer.Reset` does mention that there is no guarantee that the timer handler won't run twice. Is it safe in this case?

frostbyte73:
> Accessing the `eosTimer` struct member from the message handler seems slightly iffy from a concurrency perspective, but I get we already do in the EOS message handler... The documentation for `timer.Reset` does mention that there is no guarantee that the timer handler won't run twice. Is it safe in this case?

Yeah it's safe. I also just updated all our channels with fuses (channel + sync.Once)

biglittlebigben (approved):

PR #265, opened 2023-02-22
Restore capacity to set the segment filename suffix as a UTC timestamp

biglittlebigben:
This also updated go-glib to fix a crash.


biglittlebigben:
This still needs to be updated to use go-glib off the main branch once we merge the corresponding PR

frostbyte73 (approved):

PR #267, opened 2023-02-24
Query the S3 bucket region if not provided

biglittlebigben:
This requires appropriate permissions on the bucket

### DIFF
--- a/pkg/pipeline/sink/uploader/s3.go
+++ b/pkg/pipeline/sink/uploader/s3.go
@@ -7,10 +7,17 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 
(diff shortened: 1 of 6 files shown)


pkg/pipeline/sink/uploader/s3.go
```diff
@@ -55,7 +62,42 @@ func newS3Uploader(conf *livekit.S3Upload) Uploader {
...
+
+	if resp.LocationConstraint == nil {
+		return "", errors.ErrUploadFailed(storageFilepath, errors.New("bucket location was nil"))
+	}
+
+	return *resp.LocationConstraint, nil
+}
+
 func (u *S3Uploader) Upload(localFilepath, storageFilepath string, outputType types.OutputType) (string, int64, error) {
+	if u.awsConfig.Region == nil {
```
frostbyte73, line 91: this should happen in newS3Uploader

biglittlebigben: I was trying to avoid doing blocking I/O in the "new" routine, particularly since AWS will charge the owner of the bucket for the request. Any reason to make it early?

frostbyte73: to fail before the egress runs instead of after

frostbyte73 (approved):

PR #276, opened 2023-03-02
Fix streaming stopped errors when rtmp output fails

frostbyte73:


biglittlebigben (approved):
LGTM. Could you add some description of the pipeline errors this fixes in the commit message?

PR #279, opened 2023-03-02
Unique output queue names

frostbyte73:


pkg/pipeline/output/base.go
```diff
@@ -1,22 +1,25 @@
...
 	videoQueue *gst.Element
 }
 
-func (b *Bin) buildOutputBase(p *config.PipelineConfig) (*outputBase, error) {
+func (b *Bin) buildOutputBase(p *config.PipelineConfig, egressType types.EgressType) (*outputBase, error) {
 	base := &outputBase{}
 
 	if p.AudioEnabled {
-		audioQueue, err := builder.BuildQueueWithLatency("audio_out_queue", p.Latency, true)
+		audioQueue, err := builder.BuildQueueWithLatency(fmt.Sprintf("audio_%s_queue", egressType), p.Latency, true)
```
davidzhao, line 22: is this for debugging ? or that they have to be unique?

frostbyte73: just for debugging, gstreamer will just do queue0, queue1, etc

davidzhao (approved):

PR #281, opened 2023-03-04
Disable async-finalize on the splitmuxsink for segmented output

biglittlebigben:
This works around what seems like a race in the GStreamer splitmuxsink where the sink would sometimes send a fragment-closed event with a 0 run-time, causing the egress to fail.


frostbyte73 (approved):

frostbyte73:
`if timer := p.eosTimer; timer != nil {
	timer.Reset(eosTimeout)
}`
can also be removed from handleMessageElement case msgFragmentOpened and case msgFragmentClosed

biglittlebigben:
> `if timer := p.eosTimer; timer != nil { timer.Reset(eosTimeout) }`
> can also be removed from handleMessageElement case msgFragmentOpened and case msgFragmentClosed

We should be able to remove them indeed. I'd rather wait to make sure the pipeline behaves as we think with async-finalize disabled though

PR #284, opened 2023-03-07
Fix segment filenames

frostbyte73:


biglittlebigben (approved):
LGTM. Should we add some test cases for non trivial path combinations in the integration tests?

PR #305, opened 2023-03-22
Sdk timestamping

frostbyte73:


pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -354,38 +385,46 @@ func (w *AppWriter) pushBlankFrame(timestamp uint32) error {
...
-		if w.lastRTP > wrapCheck && pkt.Timestamp < wrapCheck {
-			w.rtpWrap++
-		}
-		w.lastRTP = pkt.Timestamp
-
 		if !blankFrame {
 			// update sequence number
 			pkt.SequenceNumber += w.snOffset
+
+			// record max frame size
```
boks1971, line 392: maybe a better name is `frameDuration` as size might be interpreted as number of bytes consumed by the largest frame.

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -354,38 +385,46 @@ func (w *AppWriter) pushBlankFrame(timestamp uint32) error {
...
+
+		// get PTS
 		pts, err := w.getPTS(pkt.Timestamp)
 		if err != nil {
 			return err
 		}
 
+		// record timing info
+		w.lastSN = pkt.SequenceNumber
+		w.lastTS = pkt.Timestamp
```
boks1971, line 416: I remember some place that ordering happens elsewhere and this is the newest values. Not sure if I am imagining things. Are these supposed to be the highest/latest values? If so, is there a need to check for out-of-order before updating these?

pkg/pipeline/source/sdk/synchronizer.go
```diff
@@ -199,70 +221,101 @@ func (t *TrackSynchronizer) onSenderReport(pkt *rtcp.SenderReport, ntpStart time
 	}
 }
 
-func absGreater(value, max time.Duration) bool {
-	return value > max || value < -max
+func (t *TrackSynchronizer) getFrameSize() uint32 {
+	if frameSize := t.frameSize; frameSize != 0 {
+		return frameSize
+	}
+	return uint32(float64(t.clockRate) / 2.4)
```
boks1971, line 228: why 2.4? Would be good to have some comment in code explaining that.

boks1971 (approved):
lgtm!

pkg/pipeline/source/sdk/appwriter.go
frostbyte73: yeah, the sample builder acts as a jitter buffer, and this happens after the packets have gone through that

pkg/pipeline/source/sdk/synchronizer.go
frostbyte73: that was supposed to be 24, good catch!

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -354,38 +385,46 @@ func (w *AppWriter) pushBlankFrame(timestamp uint32) error {
...
-		}
-		w.lastRTP = pkt.Timestamp
-
 		if !blankFrame {
 			// update sequence number
 			pkt.SequenceNumber += w.snOffset
+
+			// record max frame duration
+			if w.lastTS != 0 && pkt.SequenceNumber == w.lastSN+1 {
+				if frameDurationRTP := pkt.Timestamp - w.lastTS; frameDurationRTP > w.frameDurationRTP && frameDurationRTP < w.clockRate/60 {
```
biglittlebigben, line 394: Just curious about the `w.clockRate/60` test here?

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -354,38 +385,46 @@ func (w *AppWriter) pushBlankFrame(timestamp uint32) error {
...
+					w.frameDurationRTP = frameDurationRTP
+				}
+			}
+
 			w.translatePacket(pkt)
 		}
 
-		// will return io.EOF if EOS has been sent
+		// reset offsets if needed
+		if w.lastTS != 0 && pkt.SequenceNumber-w.lastSN > maxDropout && w.lastSN-pkt.SequenceNumber > maxDropout {
```
biglittlebigben, line 403: Should the last `&&` be a `||` here? If not, can the test ever be true?

frostbyte73: these are uint32, so the && is checking that the absolute value of the difference is > maxDropout

pkg/pipeline/source/sdk/appwriter.go
frostbyte73: That actually isn't supposed to be there - the goal was to make sure it's reasonable, will change to `< w.clockRate` to keep it < 1 second

PR #328, opened 2023-04-10
Clean filepaths

frostbyte73:


pkg/config/output.go
```diff
@@ -577,3 +577,11 @@ func redactStreamKeys(stream *livekit.StreamOutput) {
 		}
 	}
 }
+
+func clean(filepath string) string {
```
biglittlebigben, line 581: It looks like `path.Clean` will not remove `../` at the beginning of a relative path (it will for an absolute path, such as `/..`). Unless we are already enforcing this somewhere else, should be make sure that the resulting output doesn't start with `..`?

pkg/config/output.go
```diff
@@ -579,9 +579,14 @@ func redactStreamKeys(stream *livekit.StreamOutput) {
 }
 
 func clean(filepath string) string {
+	if filepath == "" {
+		return filepath
+	}
+	for strings.HasPrefix(filepath, "../") {
```
biglittlebigben, line 585: I think this should be done after the call to `path.Clean`. For instance in this case: `foo/../../bar` will end up as `../bar` after the call to `path.Clean`.

biglittlebigben (approved):

PR #331, opened 2023-04-11
Try to find default codec that is compatible with all output 

biglittlebigben:


pkg/config/output.go
```diff
@@ -211,28 +211,18 @@ func (p *PipelineConfig) updateDirectOutput(req *livekit.TrackEgressRequest) err
...
-			} else {
-				outputType = types.OutputTypeMP4
-			}
-		case livekit.EncodedFileType_MP4:
-			outputType = types.OutputTypeMP4
-		case livekit.EncodedFileType_OGG:
+	switch file.FileType {
+	case livekit.EncodedFileType_DEFAULT_FILETYPE:
+		if !p.VideoEnabled && p.AudioOutCodec != types.MimeTypeAAC {
 			outputType = types.OutputTypeOGG
```
frostbyte73, line 218: would this work for audio only file + segments?

frostbyte73 (approved):

PR #337, opened 2023-04-14
Limit dimensions

frostbyte73:


davidzhao (approved):

biglittlebigben (commented):
For ingress, I avoided limiting the resolution on the OSS side of things, but implemented it only for cloud. The logic here is that there is no reason to limit the dimensions on a custom deployment. 

On Ingress, I implemented the enforcement in cloud-io. It's probably not doable for egress as there is no synchronous call to cloud-io in the start path, but it should be possible in the egress service in cloud.

Alternatively, this could be a conf entry for egress.

PR #348, opened 2023-04-25
fix panics on zero hls segment size and concurrent ws writes

paulwe:


pkg/pipeline/sink/segments.go
```diff
@@ -55,6 +55,10 @@ func newSegmentSink(u uploader.Uploader, conf *config.PipelineConfig, p *config.
 	if maxDuration == 0 {
 		maxDuration = time.Hour * 48
 	}
+	segmentDuration := p.SegmentDuration
```
frostbyte73, line 58: handled in pkg/config/output, should default to either KeyFrameInterval or 4

frostbyte73: was there actually a panic here? That would mean something is broken in the config

frostbyte73 (approved):

PR #355, opened 2023-05-03
Replace grafov/m3u8

frostbyte73:


pkg/pipeline/sink/m3u8/reader.go
```diff
@@ -0,0 +1,62 @@
...
+func ReadPlaylist(filename string) (*Playlist, error) {
+	b, err := os.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	lines := strings.Split(string(b), "\n")
+	version, _ := strconv.Atoi(strings.Split(lines[1], ":")[1])
+	mediaType := strings.Split(lines[2], ":")[1]
+	targetDuration, _ := strconv.Atoi(strings.Split(lines[5], ":")[1])
```
biglittlebigben, line 33: This will only work with playlists we generated as it assumes the order of the tags. Should we either put a disclaimer at the top of the file, or switch to using regular expression to be more compatible?

pkg/pipeline/sink/m3u8/reader.go
```diff
@@ -0,0 +1,62 @@
...
+
+	p := &Playlist{
+		Version:        version,
+		MediaType:      mediaType,
+		TargetDuration: targetDuration,
+		Segments:       make([]*Segment, 0),
+	}
+
+	for i := 6; i < len(lines)-3; i += 3 {
+		startTime, _ := time.Parse(time.RFC3339Nano, strings.SplitN(lines[i], ":", 2)[1])
```
biglittlebigben, line 43: This also assumes the PROGRAM-DATE-TIME tag is present, which may not be true in the general case.

pkg/pipeline/sink/m3u8/writer.go
```diff
@@ -0,0 +1,75 @@
...
+func (p *PlaylistWriter) Append(dateTime time.Time, duration float64, filename string) error {
+	f, err := os.OpenFile(p.filename, os.O_WRONLY|os.O_APPEND, fs.ModeAppend)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	var sb strings.Builder
+	sb.WriteString("#EXT-X-PROGRAM-DATE-TIME:")
+	sb.WriteString(dateTime.UTC().Format(time.RFC3339Nano))
```
biglittlebigben, line 54: The spec mentions SO/IEC 8601:2004 as the spec for the date which is almost the same, but with millisecond accuracy instead of nanosecond. Using nanosecond may break some parsers.

biglittlebigben (commented):
Thanks for doing this. One small spec compliance remark. I also think that we should either make the reader a little bit more compliant, or move it to the test package to make it clear it's not general purpose.

biglittlebigben (approved):

PR #371, opened 2023-05-24
Appsrc state

frostbyte73:


biglittlebigben (approved):

davidzhao (approved):
change looks good, though it's not clear what the intended effect is. Could you add more to the description so others could use it as a reference please?

PR #374, opened 2023-06-01
Identifiable redacted keys

frostbyte73:
Stream keys will show up as `{123...XYZ}` instead of `**********`.
Access keys and secrets will be named, for example `{access_key}`

### DIFF
--- a/pkg/config/output.go
+++ b/pkg/config/output.go
@@ -219,35 +219,35 @@ func redactEncodedOutputs(out EncodedOutput) {
 	}
 }
 
-func redactUpload(upload uploader) {
-	if s3 := upload.GetS3(); s3 != nil {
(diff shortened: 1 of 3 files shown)


noahlt (approved):
LGTM

I wonder if it would be slightly more legible (and future-proof) if stream keys were `{stream_key:abc...123}`. Not a blocker though

biglittlebigben (approved):

frostbyte73:
> LGTM
> 
> I wonder if it would be slightly more legible (and future-proof) if stream keys were `{stream_key:abc...123}`. Not a blocker though

I was trying to find a way to make it readable, look good, and also be kept short for display purposes. But I think anyone using rtmp output will know what that is without the name/identifier

PR #378, opened 2023-06-01
AwaitStartSignal web option

frostbyte73:


pkg/config/pipeline.go
```diff
@@ -161,6 +162,7 @@ func (p *PipelineConfig) Update(request *rpc.StartEgressRequest) error {
 		redactEncodedOutputs(clone)
 
 		p.SourceType = types.SourceTypeWeb
+		p.AwaitStartSignal = true
```
biglittlebigben, line 165: Not really a comment on this PR, but rather about the matching protocol change: I think the semantics of this options are unclear as they stand. We should make it more clear that this only applied to web egress.

biglittlebigben (approved):

PR #381, opened 2023-06-07
send update after stream updates

frostbyte73:


pkg/pipeline/pipeline.go
```diff
@@ -280,19 +281,26 @@ func (p *Pipeline) UpdateStream(ctx context.Context, req *livekit.UpdateStreamRe
...
 	for _, url := range req.RemoveOutputUrls {
-		if err := p.removeSink(url, nil); err != nil {
+		if err := p.removeSink(ctx, url, nil); err != nil {
 			errs.AppendErr(err)
+		} else {
+			sendUpdate = true
 		}
 	}
 
+	if sendUpdate && p.onStatusUpdate != nil {
```
davidzhao, line 296: This is when a user adds a new stream right? Does it also send an update when a particular stream fails? From the test it seems to indicate that it does happen. I'm just not able to spot where.

davidzhao (approved):

biglittlebigben (approved):

pkg/pipeline/pipeline.go
frostbyte73: `removeSink` calls 
```
	// only send updates if the egress will continue, otherwise it's handled by UpdateStream RPC
	if streamErr != nil {
		p.Info.UpdatedAt = time.Now().UnixNano()
		p.sendUpdate(ctx, p.Info)
	}
```

PR #399, opened 2023-06-26
Mark remaining streams as failed if egress fails

frostbyte73:


davidzhao:
Question: when does the outer Egress fail? Would this accidentally override other streams that have succeeded?

frostbyte73:
> Question: when does the outer Egress fail? Would this accidentally override other streams that have succeeded?

It's rare, the example given was an ERR_CONNECTION_REFUSED failure to load the website for a web egress. That map object contains only active streams, so any streams that already finished won't be changed

davidzhao (approved):

PR #403, opened 2023-06-28
Stop race between multiple requests

frostbyte73:


pkg/service/manager.go
```diff
@@ -138,13 +139,12 @@ func (s *ProcessManager) awaitCleanup(h *process) {
 		s.onFatalError(h.info)
 	}
 
-	h.closed.Break()
 	s.monitor.EgressEnded(h.req)
+	h.closed.Break()
 
 	s.mu.Lock()
-	defer s.mu.Unlock()
```
davidzhao, line 145: in this case defer seems fine?

pkg/stats/monitor.go
```diff
@@ -146,17 +151,15 @@ func (m *Monitor) GetCPULoad() float64 {
...
-	if total-available <= 0.01 {
-		// if idle, add room for when cost == numCPUs
-		available += 0.01
-	} else {
-		// if not idle, leave 20% CPU
-		available -= total * 0.2
+	count := m.egressCount.Load()
+	reserved := m.pendingCPUs.Load()
+	if count > 0 {
+		reserved += (numCPU - m.cpuStats.GetCPUIdle()) * math.Pow(breathingRoom, float64(count))
```
davidzhao, line 159: what's the intention of this? I'm plugging some numbers and it doesn't seem to produce sane values for me. can we add tests for this logic to demonstrate the expected & edge cases?

i.e. 8 cpus, with 4 idle and 2 current egresses.. the logic becomes:

(8-4) * 1.2^2 = 5.76 reserved

also should all egress types be treated the same? what happens when I have 20 track egresses that consume no CPU.. wouldn't the reserved count be out of wack?

frostbyte73: the idea was the more egresses that are running, the more of a risk it is to add another. So 8 cpus with 4 idle and 2 egresses -> 5.76 reserved, 3.24 remaining, if cpu cost is 3 it would accept a third

8 cpus with 3 idle and 5 egresses -> (8-3) * 1.2^5 = 12.4, and it wouldn't accept a 6th

pkg/stats/monitor.go
```diff
@@ -171,45 +188,80 @@ func (m *Monitor) CanAcceptRequest(req *rpc.StartEgressRequest) bool {
 	return accept
 }
 
-func (m *Monitor) AcceptRequest(req *rpc.StartEgressRequest) {
+func (m *Monitor) AcceptRequest(req *rpc.StartEgressRequest) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.CanAcceptRequest(req) {
```
davidzhao, line 195:
```suggestion
	if !m.canAcceptRequest(req) {
```

already under lock

pkg/stats/monitor.go
```diff
@@ -171,45 +188,80 @@ func (m *Monitor) CanAcceptRequest(req *rpc.StartEgressRequest) bool {
...
+	}
+}
+
+func (m *Monitor) EgressAborted(req *rpc.StartEgressRequest) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	switch req.Request.(type) {
 	case *rpc.StartEgressRequest_RoomComposite:
-		m.requestGauge.With(prometheus.Labels{"type": "room_composite"}).Sub(1)
```
davidzhao, line 207: we don't need to subtract from gauge?

davidzhao (approved):
this seems cleaner. lgtm

pkg/stats/monitor.go
frostbyte73: this is called when something fails when EgressStarted hasn't been called yet, but the request had already been accepted

PR #405, opened 2023-06-29
Delete segments after upload if upload is enabled

biglittlebigben:
Is this safe to do?


frostbyte73:
`*Uploader` is the uploader base, it should clean up in uploader.go L61.
We'll also need to add a parameter to upload, because we don't want to delete the playlist after uploading but everything else should be ok

pkg/pipeline/sink/uploader/uploader.go
```diff
@@ -89,6 +94,17 @@ func (u *noOpUploader) upload(localFilepath, _ string, _ types.OutputType) (stri
...
 }
 
+func (u *noOpUploader) cleanupFile(localFilepath string) error {
+	return nil
+}
+
+type baseUploader struct {
+}
+
+func (u *baseUploader) cleanupFile(localFilepath string) error {
```
frostbyte73, line 104: `func (u *Uploader) cleanupFile(localFilepath string) error {`

biglittlebigben: Wouldn't the cause the file to be deleted as well in cases where there is no cloud upload (which I believe means that the noopUploader is selected)?

frostbyte73: oh, true. In that case, better to kill `*Uploader` and change the `uploader` interface to be public. 
Then the backup logic can move into `baseUploader`

frostbyte73 (approved):

PR #414, opened 2023-07-06
Chrome sandbox

frostbyte73:


pkg/pipeline/source/web.go
```diff
@@ -130,7 +130,9 @@ func (s *WebSource) createPulseSink(ctx context.Context, p *config.PipelineConfi
 		return errors.Fatal(errors.ErrProcessStartFailed(err))
 	}
 
-	s.pulseSink = strings.TrimRight(b.String(), "\n")
+	out := b.String()
+	logger.Infow("pulse", "output", out)
```
davidzhao, line 134: intended to be info?

test/room_composite.go
```diff
@@ -36,6 +36,9 @@ func (r *Runner) runRoomTest(t *testing.T, name string, audioCodec, videoCodec t
 }
 
 func (r *Runner) testRoomCompositeFile(t *testing.T) {
+	time.Sleep(time.Minute * 10)
```
davidzhao, line 39: 🤔

davidzhao (approved):

pkg/pipeline/source/web.go
frostbyte73: going to be removed

PR #417, opened 2023-07-07
Remove local_directory

frostbyte73:
With sandboxing/no longer running as root, the user can only create folders under /home/egress


davidzhao (approved):
this will require a minor version bump? since we are getting rid of this config option

PR #427, opened 2023-07-11
Close websocketsink on read closeError

frostbyte73:


pkg/pipeline/sink/websocket.go
```diff
@@ -53,6 +55,11 @@ func (s *WebsocketSink) Start() error {
 		for {
 			_, _, err := s.conn.ReadMessage()
 			if err != nil {
+				if _, ok := err.(*websocket.CloseError); ok {
```
davidzhao, line 58: it'd be good to check against a range of diff close errors like we do [here](https://github.com/livekit/livekit/blob/master/pkg/service/rtcservice.go#L351)

davidzhao (approved):

PR #472, opened 2023-08-24
Video selector

frostbyte73:


pkg/pipeline/source/sdk.go
```diff
@@ -145,78 +145,94 @@ func (s *SDKSource) joinRoom() error {
...
+	return nil
+}
 
-	if err := s.UpdateInfoFromSDK(fileIdentifier, s.filenameReplacements); err != nil {
-		logger.Errorw("could not update file params", err)
+func (s *SDKSource) awaitTracks(expecting map[string]struct{}) error {
+	trackCount := len(expecting)
+	s.errors = make(chan error, trackCount)
+
+	deadline := time.After(subscriptionTimeout)
```
biglittlebigben, line 190: nit: would a context be a more self contained alternative here?

biglittlebigben (approved):

pkg/pipeline/source/sdk.go
frostbyte73: will move that over with participant composite

PR #474, opened 2023-08-25
Participant Composite

frostbyte73:


biglittlebigben (approved):
Great! I think we may just consider relying a bit more on contexts for some of the cancellation/timeout handling.

pkg/gstreamer/bin.go
```diff
@@ -253,7 +257,19 @@ func (b *Bin) RemoveSinkBin(name string) (bool, error) {
 }
 
 func (b *Bin) SetState(state gst.State) error {
-	return b.bin.SetState(state)
+	stateErr := make(chan error, 1)
+	go func() {
+		stateErr <- b.bin.SetState(state)
+	}()
+	select {
+	case <-time.After(stateChangeTimeout):
```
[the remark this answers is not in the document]
frostbyte73: if you're passing the context into other functions/doing more with deadlines I'd agree, but for simple cases like this I prefer the simplicity of time.After. One line less without the `defer cancel()`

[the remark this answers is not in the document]
biglittlebigben: Indeed. My point would be that I think that ideal `(b *Bin) SetState` (and potentially callers as well) should take a context as well, allowing the caller to cancel the wait on timeout.

Retrofitting all this is beyond the scope of this PR, of course.

pkg/pipeline/source/sdk.go
```diff
@@ -326,16 +380,21 @@ func (s *SDKSource) onTrackSubscribed(track *webrtc.TrackRemote, pub *lksdk.Remo
 	if !s.initialized.IsBroken() {
 		s.mu.Lock()
 		switch s.RequestType {
+		case types.RequestTypeParticipant:
+			s.filenameReplacements["{publisher_identity}"] = s.Identity
```
davidzhao, line 384: would we make pID a replacement policy too? not sure if that needs to be onTrackSubscribed tho.

davidzhao (approved):
nice!

pkg/pipeline/source/sdk.go
frostbyte73: could add ID if/when someone asks, I think for the most part identity is more useful

davidzhao: yeah you are right.. this is participant composite.. identity is the main thing people would care about.

PR #476, opened 2023-08-30
Add support for live playlists in SegmentedFileOutput

biglittlebigben:


biglittlebigben:
This hardcodes the live playlist window to 5 segments. We could expose this in the protocol, but I'm not sure there is any reason to overload the request until we get a use case.

frostbyte73 (approved):

PR #506, opened 2023-09-28
Support for Image output in egress

biglittlebigben:
This adds preliminary support for generating images at regular interval for a session in egress. This is still mostly untested with only basic integration test support.

### DIFF
--- a/pkg/pipeline/source/sdk.go
+++ b/pkg/pipeline/source/sdk.go
@@ -167,10 +167,11 @@ func (s *SDKSource) joinRoom() error {
 
 	var fileIdentifier string
 	var err error
+	var w, h uint32
 	switch s.RequestType {
 	case types.RequestTypeParticipant:
 		fileIdentifier = s.Identity
-		err = s.awaitParticipant(s.Identity)
+		w, h, err = s.awaitParticipant(s.Identity)
 
 	case types.RequestTypeTrackComposite:
 		fileIdentifier = s.Info.RoomName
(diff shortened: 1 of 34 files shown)


pkg/pipeline/builder/stream.go
```diff
@@ -98,6 +98,12 @@ func BuildStreamBin(pipeline *gstreamer.Pipeline, p *config.PipelineConfig) (*St
 		}
 	}
 
+	b.SetGetSrcPad(func(name string) *gst.Pad {
+		fmt.Println("NAME", name)
```
frostbyte73, line 102:
```suggestion
```

frostbyte73, line 102: debug print

pkg/pipeline/builder/websocket.go
```diff
@@ -33,5 +34,9 @@ func BuildWebsocketBin(pipeline *gstreamer.Pipeline, appSinkCallbacks *app.SinkC
 		return nil, err
 	}
 
+	b.SetGetSrcPad(func(name string) *gst.Pad {
```
frostbyte73, line 37: this probably isn't needed since the sink will only have one template

pkg/pipeline/source/sdk.go
```diff
@@ -181,40 +182,49 @@ func (s *SDKSource) joinRoom() error {
...
 
 	for trackCount := 0; trackCount == 0 || trackCount < len(rp.Tracks()); trackCount++ {
 		if err = <-s.errors; err != nil {
-			return err
+			return 0, 0, err
+		}
+	}
+
+	var w, h uint32
+	for _, t := range rp.Tracks() {
```
frostbyte73, line 218: what happens if the participant publishes video late?

pkg/pipeline/builder/websocket.go
biglittlebigben: Removed

pkg/pipeline/source/sdk.go
biglittlebigben: Currently, it will default to the video encoding settings (which themselves default to 1080p). I just added a small fix to make sure this is true.

Changing the behavior to start the image output only when a video track joins would be better, but I do not have any way to add the image output on the running pipeline yet. This will come when I implement "UpdateOutputs".

frostbyte73 (approved):

PR #508, opened 2023-09-29
force keyframe interval for segments

frostbyte73:


pkg/config/output.go
```diff
@@ -148,6 +148,14 @@ func (p *PipelineConfig) updateEncodedOutputs(req EncodedOutput) error {
 		}
 	}
 
+	if segmentConf := p.Outputs[types.EgressTypeSegments]; segmentConf != nil {
+		// double the segment length
+		p.KeyFrameInterval = float64(2 * segmentConf.(*SegmentConfig).SegmentDuration)
```
biglittlebigben, line 153: Where do we set the default segment duration and what is it set to?

frostbyte73: it's set to 4 in the output_segments

biglittlebigben (approved):

PR #132, opened 2022-10-13
Fix starting stopped container

nkonev:
Prior this commit it's impossible to start again already stopped container.
```
[nkonev@fedora ~]$ docker logs -f videochat_egress_1
+ export TMPDIR=/tmp/lkegress
+ TMPDIR=/tmp/lkegress
+ pulseaudio -D --verbose --exit-idle-time=-1 --system --disallow-exit
W: [pulseaudio] main.c: Running in system mode, but --disallow-module-loading not set.
N: [pulseaudio] main.c: Running in system mode, forcibly disabling SHM mode.
I: [pulseaudio] main.c: Daemon startup successful.
+ '[' -z /tmp/lkegress ']'
+ mkdir -p /tmp/lkegress
+ exec egress
{"level":"info","ts":1665693000.0708337,"logger":"egress","caller":"redis/redis.go:53","msg":"connecting to redis","nodeID":"NE_wQ5UGX87HS5d","sentinel":false,"addr":"redis:6379"}

^C
[nkonev@fedora ~]$ docker stop videochat_egress_1
videochat_egress_1
[nkonev@fedora ~]$ docker start videochat_egress_1
videochat_egress_1
[nkonev@fedora ~]$ docker logs -f videochat_egress_1
+ export TMPDIR=/tmp/lkegress
+ TMPDIR=/tmp/lkegress
+ pulseaudio -D --verbose --exit-idle-time=-1 --system --disallow-exit
W: [pulseaudio] main.c: Running in system mode, but --disallow-module-loading not set.
N: [pulseaudio] main.c: Running in system mode, forcibly disabling SHM mode.
I: [pulseaudio] main.c: Daemon startup successful.
+ '[' -z /tmp/lkegress ']'
+ mkdir -p /tmp/lkegress
+ exec egress
{"level":"info","ts":1665693000.0708337,"logger":"egress","caller":"redis/redis.go:53","msg":"connecting to redis","nodeID":"NE_wQ5UGX87HS5d","sentinel":false,"addr":"redis:6379"}
{"level":"info","ts":1665693008.62964,"logger":"egress","caller":"server/main.go:100","msg":"exit requested, finishing recording then shutting down","nodeID":"NE_wQ5UGX87HS5d","signal":"terminated"}
{"level":"info","ts":1665693008.6298237,"logger":"egress","caller":"service/service.go:98","msg":"shutting down","nodeID":"NE_wQ5UGX87HS5d"}
+ export TMPDIR=/tmp/lkegress
[...]

PR #511, opened 2023-10-16
Fix start after stopping

nkonev:
Hi!
I noticed that start-after-stopping isn't working in the latest version v1.7.12. It seems it caused by 331c179a.

PTAL @frostbyte73

In general, it follows the same direction as my previous PR https://github.com/livekit/egress/pull/132


nkonev:
test failure looks unrelated, if I understood correctly, Github does not pass `EGRESS_CONFIG_STRING="$(echo ${{ secrets.EGRESS_CONFIG_STRING }} | base64 -d)"`m for builds, triggered by me

cc @biglittlebigben @frostbyte73

frostbyte73 (approved):

PR #521, opened 2023-10-26
Add more options to the S3 section of config.yaml for egress

dandoug:
Added additional S3 configuration to the `config.yaml` for egress to allow specification of

* HTTP proxy
* max retries
* max/min retry delay
* AWS logging option

The change does not alter the `livekit.S3Upload` struct which is shared across components.  It extends it within the egress component to add the additional information.   The intended usage is in an environment where egress files on requests would NOT specify the S3 file information.... but if such info was specified, the information in the extended part of the s3 `config.yaml` would be added (uploads.go lines 48-56).

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,12 @@ s3:
   region: AWS_DEFAULT_REGION env or IAM role can be used instead
   endpoint: (optional) custom endpoint
   bucket: bucket to upload files to
+  # the following s3 options can only be set in config, *not* per request, they will be added to any per-request options
+  proxy: (optional, no default) proxy url
+  max_retries: (optional, default=3) number or retries to attempt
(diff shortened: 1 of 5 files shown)


dandoug:
sorry for the force pushes... I tried to move this to be based on 1.8.0, but that requires a version of livekit-server not yet available so I moved back to be based on 1.7.12

dandoug:
btw... I reviewed the CLA with the eBay Open Source counsel this morning and he confirmed that we had no issues with it as long as the PR did not contain eBay propriety information or patentable IP.  I assured him that it did not.

frostbyte73:
Closing in favor of #529. We're releasing livekit server v1.5.1 today so you should be able to move to 1.8

PR #522, opened 2023-10-26
Update stream reset logic

frostbyte73:


pkg/pipeline/builder/stream.go
```diff
@@ -210,22 +211,21 @@ func (sb *StreamBin) MaybeResetStream(name string, streamErr error) (bool, error
 	}
 	values := s.(*gst.Structure).Values()
 	outBytes := values["out-bytes-acked"].(uint64)
-	if outBytes == 0 {
+
+	if sink.reconnections == 0 && outBytes == 0 {
```
davidzhao, line 215: would this timeout after the first retry had failed? i.e. we would try to start the pipeline and `outBytes` would remain 0

frostbyte73: `outBytes` gets reset

biglittlebigben (approved):

PR #529, opened 2023-11-07
apply changes from handler-metrics to egress 1.8.0 base

dandoug:
These are the same changes from #526 but 

- I've cherry-picked them
- added the changes to `image.go`, `image_manifest.go`, and `sink.go` that I could not at the 1.7.12 level
- made the minor changes in Dockerfile and go.mod as indicated by the build

I'm unable to build and test these in our environment due to running an older `livekit-server` but the changes should be the same.

cc: @frostbyte73 @biglittlebigben


dandoug:
@biglittlebigben @frostbyte73 
So, the integeration test here failed
```
+ /home/egress/go/bin/gotestfmt
📦 egress
❌ TestEgress (0s)
      runner.go:76: 
          	Error Trace:	/workspace/test/runner.go:76
          	            				/workspace/test/integration_test.go:38
          	Error:      	Should NOT be empty, but was 
          	Test:       	TestEgress

Error: Process completed with exit code 1.
```
Because the `EGRESS_CONFIG_STRING` env var was empty and the `EGRESS_CONFIG_FILE` was not specified either.  @frostbyte73 explained to me before that that was because the config contains creds you don't want exposed here.  

But this PR (#529) has the metrics changes and updates for `Uploader` interface changes as requested.

dandoug:
FYi... here's a sample panel created with the added handler metrics from egress in this PR

egressMetricsSample

pkg/pipeline/sink/file.go
```diff
@@ -44,10 +43,8 @@ func (s *FileSink) Start() error {
 }
 
 func (s *FileSink) Close() error {
-	location, size, err := s.Upload(s.LocalFilepath, s.StorageFilepath, s.OutputType, false)
-	if err != nil {
-		return err
-	}
+	location, size, err := s.Upload(s.LocalFilepath, s.StorageFilepath, s.OutputType, false, "file")
+
```
frostbyte73, line 47: this error needs to be returned at some point

pkg/pipeline/sink/uploader/s3.go
```diff
@@ -34,18 +37,42 @@ const (
...
 
+// CustomRetryer wraps the SDK's built in DefaultRetryer adding additional
+// custom features. Namely, to always retry.
+type CustomRetryer struct {
+	client.DefaultRetryer
+}
+
+// ShouldRetry overrides the SDK's built in DefaultRetryer because the PUTs for segments/playlists are always idempotent
+func (r CustomRetryer) ShouldRetry(req *request.Request) bool {
+	return true
```
frostbyte73, line 48: I think this still needs some safeguards - in the case of bad credentials (which happens fairly often in cloud), it will take much longer to fail

pkg/pipeline/sink/uploader/s3.go
```diff
@@ -34,18 +37,42 @@ const (
...
+				MaxRetryDelay:    conf.MaxRetryDelay,
+				MaxThrottleDelay: conf.MaxRetryDelay,
+				MinRetryDelay:    conf.MinRetryDelay,
+				MinThrottleDelay: conf.MinRetryDelay,
+			},
+		},
 		S3ForcePathStyle: aws.Bool(conf.ForcePathStyle),
+		LogLevel:         aws.LogLevel(conf.AwsLogLevel),
 	}
+	logger.Infow("setting AWS config", "maxRetries", conf.MaxRetries,
```
frostbyte73, line 72: nit: debug level

pkg/service/handler.go
```diff
@@ -196,6 +201,47 @@ func (h *Handler) GetPProf(ctx context.Context, req *ipc.PProfRequest) (*ipc.PPr
...
+	}
+
+	logger.Debugw("returning metrics from handler process", "sizeOfFamilies", len(metrics))
+	metricsAsString, cnt, err := renderMetrics(metrics)
+	if err != nil {
+		return &ipc.MetricsResponse{
+			Metrics: "",
+		}, err
+	}
+	logger.Debugw("Metrics returned from handler process", "cnt", cnt, "metrics", metricsAsString)
```
frostbyte73, line 221: nit: lower case log messages here and below in renderMetrics

frostbyte73 (changes requested):
Looks good overall, just a few small changes.
Can you also sort the imports with goimports?

pkg/pipeline/sink/file.go
dandoug: fixed so the error is returned

pkg/pipeline/sink/uploader/s3.go
dandoug: added check for HTTP status codes 401 or 403 and does not retry for those

pkg/pipeline/sink/uploader/s3.go
dandoug: changed to debug

pkg/pipeline/sink/segments.go
```diff
@@ -139,11 +153,7 @@ func (s *SegmentSink) handleClosedSegment(update SegmentUpdate) {
 	go func() {
 		defer close(update.uploadComplete)
 
-		_, size, err := s.Upload(segmentLocalPath, segmentStoragePath, s.outputType, true)
-		if err != nil {
-			s.callbacks.OnError(err)
-			return
-		}
+		_, size, _ := s.Upload(segmentLocalPath, segmentStoragePath, s.outputType, true, "segment")
```
frostbyte73, line 156: I missed this one earlier, in the case of bad credentials the egress should fail

pkg/pipeline/sink/uploader/s3.go
```diff
@@ -71,6 +98,22 @@ func newS3Uploader(conf *livekit.S3Upload) (uploader, error) {
 		u.awsConfig.Region = aws.String(region)
 	}
 
+	if conf.Proxy != "" {
+		logger.Infow("configuring s3 with proxy", "proxyEndpoint", conf.Proxy)
```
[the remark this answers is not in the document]
dandoug: changed to debug

pkg/service/handler.go
dandoug: made message start with lower case (and below as well)

dandoug:
@frostbyte73 thanks for the review.  I made the changes you requested.

pkg/pipeline/sink/segments.go
dandoug: oh... ok.  I'll add
```
if err != nil {
			s.callbacks.OnError(err)
			return
		}
```

dandoug:
@frostbyte73 added segment Upload error handler per your latest review comment.

frostbyte73 (approved):

PR #538, opened 2023-11-14
Replace yarn with pnpm

lukasIO:
... and update to node18 for docker and gh actions


davidzhao (approved):
thanks for taking care of this!

PR #539, opened 2023-11-14
Uploads cleanup

frostbyte73:


test/download.go
```diff
@@ -35,25 +35,28 @@ import (
...
 
+	"github.com/livekit/egress/pkg/config"
 	"github.com/livekit/protocol/livekit"
 	"github.com/livekit/protocol/logger"
 )
 
 func download(t *testing.T, uploadParams interface{}, localFilepath, storageFilepath string) {
-	logger.Debugw("download", "localFilepath", localFilepath, "storageFilepath", storageFilepath)
 	switch u := uploadParams.(type) {
-	case *livekit.S3Upload:
```
davidzhao, line 45: do we need both of them here?

davidzhao (approved):

test/download.go
frostbyte73: EgressS3Upload replaces the old one

PR #541, opened 2023-11-18
Eos timer

frostbyte73:


pkg/pipeline/watch.go
```diff
@@ -121,6 +121,7 @@ func (c *Controller) messageWatch(msg *gst.Message) bool {
 	switch msg.Type() {
 	case gst.MessageEOS:
 		logger.Infow("EOS received")
+		c.eosTimer.Stop()
```
davidzhao, line 124: nil check? possible EOS is sent by something else?

frostbyte73: shouldn't be possible but might as well add it

pkg/pipeline/controller.go
```diff
@@ -405,7 +407,12 @@ func (c *Controller) SendEOS(ctx context.Context) {
 
 		case livekit.EgressStatus_EGRESS_ENDING,
 			livekit.EgressStatus_EGRESS_LIMIT_REACHED:
-			go c.p.SendEOS()
+			go func() {
+				c.eosTimer = time.AfterFunc(time.Second*30, func() {
+					c.OnError(errors.ErrPipelineFrozen)
+				})
+				c.p.SendEOS()
```
biglittlebigben, line 414: Will this kill in process uploads? We have many examples of cases where upload takes more than 30 seconds (a 3h file could be several GB, and even for shorter sessions, we have cases where S3 is just slow). We should not charge customers for that time, but we should keep the upload going.

frostbyte73: the media won't be available if the eos is never received

frostbyte73: (unless it's HLS, in which case the playlist will not be closed)

davidzhao (approved):

pkg/pipeline/controller.go
biglittlebigben: My question is about calling `c.OnError(errors.ErrPipelineFrozen)` after 30s.

Issue #518, opened 2023-10-24
[FEATURE] Automatic attach of custom metadata when uploading recordings to Amazon S3


JSON-OBJECT:
## TLDR
  * When uploading a recording file to Amazon S3 in LiveKit Egress, is it possible to add Content-Disposition: attachment as metadata?

## Detailed explanation and questions
  * In my project, I use Kotlin SDK to make a recording request to Amazon S3 through LiveKit Egress. In this case, I am using it as below code. Is it possible to automatically include Content-Disposition: attachment in the metadata when uploading all recorded files? The reason I need this is because I want the Save as feature when downloading from the Browser. If we want to add metadata after the upload is complete, must copy the S3 Object, which is inefficient.
  
```kotlin
        val output = LivekitEgress.EncodedFileOutput.newBuilder()
            .setFileType(LivekitEgress.EncodedFileType.MP4)
            .setFilepath(".../room-composite-{time}")
            .setS3(
                LivekitEgress.S3Upload.newBuilder()
                    .setAccessKey($AWS_ACCESS_KEY_ID)
                    .setSecret($AWS_SECRET_ACCESS_KEY)
                    .setRegion($AWS_REGION)
                    .setBucket($FFL_S3_BUCKET)
                    .build()
            )
            .build()
```

PR #545, opened 2023-11-27
[S3 Upload] Expose ContentDisposition

spennihana:
Depends on https://github.com/livekit/protocol/pull/525

And is related to #518


pkg/pipeline/sink/uploader/s3.go
```diff
@@ -128,6 +129,12 @@ func newS3Uploader(conf *config.EgressS3Upload) (uploader, error) {
 
 	if conf.Tagging != "" {
 		u.tagging = aws.String(conf.Tagging)
+	} else {
+		u.tagging = aws.String("inline")
```
frostbyte73, line 133: what does `inline` do?

davidzhao: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

it does seem like attachment would be a better default.. but changing it may be disruptive to users.

frostbyte73:
@spennihana can you update the protocol reference to `fd7066503feeef328b73c7cd07414836ceb8c682`?

PR #551, opened 2023-12-01
prom can accept

frostbyte73:
Adds `can_accept_request` prometheus gauge


pkg/stats/monitor.go
```diff
@@ -74,7 +78,14 @@ func (m *Monitor) Start(conf *config.ServiceConfig, isAvailable func() float64)
 		Subsystem:   "egress",
 		Name:        "available",
```
biglittlebigben, line 79: I guess it's too late to change this to "idle"...

biglittlebigben (approved):

PR #555, opened 2023-12-07
Wait for handler to register

frostbyte73:
StartEgress will now wait for the handler to register with psrpc before returning


pkg/ipc/ipc.proto
```diff
@@ -17,6 +17,16 @@ syntax = "proto3";
 package ipc;
 option go_package = "github.com/livekit/egress/pkg/ipc";
 
+import "google/protobuf/empty.proto";
+
+service EgressService {
+  rpc HandlerReady(HandlerReadyRequest) returns (google.protobuf.Empty) {};
```
davidzhao, line 23:
```suggestion
  rpc WaitForHandlerReady(HandlerReadyRequest) returns (google.protobuf.Empty) {};
```

davidzhao (approved):

PR #566, opened 2023-12-20
per-process cpu reservations

frostbyte73:


biglittlebigben (approved):

pkg/stats/monitor.go
```diff
@@ -168,33 +188,78 @@ func (m *Monitor) GetRequestCount() int {
...
+	m.procStats[pid] = ps
+}
+
+func (m *Monitor) GetUsageStats(egressID string) (float64, float64) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	for pid, ps := range m.procStats {
+		if ps.egressID == egressID {
+			delete(m.procStats, pid)
```
davidzhao, line 212: why do we delete it here? seems like a `GetXX` should not have a side effect like this?

frostbyte73: I'll rename, it's only called after the egress ends

davidzhao (approved):

PR #568, opened 2023-12-22
CBR for rtmp streams

frostbyte73:


pkg/pipeline/builder/video.go
```diff
@@ -538,6 +535,12 @@ func (b *VideoBin) addEncoder() error {
 		if err = x264Enc.SetProperty("vbv-buf-capacity", bufCapacity); err != nil {
 			return err
 		}
+		if b.conf.GetStreamConfig() != nil {
+			x264Enc.SetArg("pass", "cbr")
```
biglittlebigben, line 539: `cbr` is the default value according to the documentation. Any reason to set it anyway?

biglittlebigben (approved):

pkg/pipeline/builder/video.go
frostbyte73: the `veryfast` setting sets `rc=crf`, I wasn't sure whether that meant it's being overridden

PR #574, opened 2024-01-03
kill egress when usage too high

frostbyte73:


pkg/stats/monitor.go
```diff
@@ -203,8 +211,30 @@ func (m *Monitor) updateEgressStats(idle float64, usage map[int]float64) {
 
 		if procStats.egressID != "" {
 			m.promProcCPULoad.With(prometheus.Labels{"egress_id": procStats.egressID}).Set(cpuUsage)
+			if cpuUsage > maxUsage {
```
davidzhao, line 214: if the maxUsage instance has not exceeded quota limits that are defined, it seems that we shouldn't kill it?

pkg/stats/monitor.go
```diff
@@ -203,8 +222,30 @@ func (m *Monitor) updateEgressStats(idle float64, usage map[int]float64) {
...
+			"load", load,
+			"requests", m.requests.Load(),
+		)
+
+		if m.requests.Load() > 1 {
+			if m.highCPUCount < 3 {
+				m.highCPUCount++
+				return
+			} else {
+				m.killProcess(maxEgress)
```
davidzhao, line 243: maxEgress could be empty? I guess that might be fine.

davidzhao (approved):

pkg/stats/monitor.go
frostbyte73: I couldn't think of a scenario where that'd be possible, but it'll be a no-op if it happens

PR #583, opened 2024-01-12
Add workflow action to create multi image manifest

biglittlebigben:
Now that we are building the gstreamer images for different architectures on different workers, each architecture image being pushed overwrites the other one, resulting in image tags with only one architecture.

In order to fix this, this PR renames the images created by each worker by appending an arch suffix to the tag. This then adds a GH action to be run manually after all images are built, to manually create a multi arch manifest that references the  arch specific tags.


davidzhao (approved):
lgtm!

frostbyte73 (approved):

PR #603, opened 2024-02-01
Add messages for aborted egress

frostbyte73:


biglittlebigben (approved):

pkg/pipeline/controller.go
```diff
@@ -215,6 +215,7 @@ func (c *Controller) Run(ctx context.Context) *livekit.EgressInfo {
 		select {
 		case <-c.stopped.Watch():
 			c.Info.Status = livekit.EgressStatus_EGRESS_ABORTED
+			c.Info.Error = "Start signal not received"
```
davidzhao, line 218: would be great if we can dump any connection error strings from the template itself.

PR #611, opened 2024-02-14
Status messaging, keyframe intervals

frostbyte73:


pkg/pipeline/controller.go
```diff
@@ -400,18 +403,14 @@ func (c *Controller) SendEOS(ctx context.Context) {
 			c.p.Stop()
 
 		case livekit.EgressStatus_EGRESS_ACTIVE:
-			c.Info.UpdatedAt = time.Now().UnixNano()
-			if c.Info.Error != "" {
-				c.Info.Status = livekit.EgressStatus_EGRESS_FAILED
-				c.p.Stop()
```
davidzhao, line 406: so we no longer log it as failed? or stop the process?

pkg/config/output.go
```diff
@@ -162,12 +162,12 @@ func (p *PipelineConfig) updateEncodedOutputs(req EncodedOutput) error {
...
 	}
 
-	if segmentConf := p.Outputs[types.EgressTypeSegments]; segmentConf != nil && len(segmentConf) > 0 {
-		// double the segment length
-		p.KeyFrameInterval = float64(2 * segmentConf[0].(*SegmentConfig).SegmentDuration)
-	} else if p.KeyFrameInterval == 0 && p.Outputs[types.EgressTypeStream] != nil {
+	if p.KeyFrameInterval == 0 && p.Outputs[types.EgressTypeStream] != nil {
 		// default 4s for streams
 		p.KeyFrameInterval = 4
+	} else if segmentConf := p.Outputs[types.EgressTypeSegments]; segmentConf != nil && len(segmentConf) > 0 {
```
biglittlebigben, line 168: We should make sure to test this change with the stream + segment and segments with a duration > 4s. In the past, I've seem issues where the splitmuxsink was generating segments of very uneven duration if the key frame interval was < the segment duration. 

Also note that splitmuxsink will force a key frame regardless of the p.KeyFrameInterval parameter when a segment split is needed.

pkg/pipeline/controller.go
frostbyte73: L401-403:
```
		case livekit.EgressStatus_EGRESS_ABORTED,
			livekit.EgressStatus_EGRESS_FAILED:
			c.p.Stop()
```

The difference is just that we now mark it as failed when setting the error instead of relying on whether or not `Error` is empty

pkg/config/output.go
frostbyte73: tested with 6s segment + stream, seems it was fixed in one of the gstreamer updates

davidzhao (approved):

PR #636, opened 2024-03-19
go 1.22

frostbyte73:


biglittlebigben (approved):
Is there anything in go 1.22 that we need?

frostbyte73:
> Is there anything in go 1.22 that we need?

no, thought it was the reason cloud-egress was failing to build but might as well

PR #643, opened 2024-03-28
fix jitter buffer panic

frostbyte73:


dennwc (approved):
CI fails because of some SFU changes. LGTM otherwise!

PR #674, opened 2024-05-15
service refactor

frostbyte73:


pkg/service/process.go
```diff
@@ -62,15 +69,150 @@ func NewProcess(
...
+	case <-time.After(launchTimeout):
+		logger.Warnw("no response from handler", nil, "egressID", info.EgressId)
+		_ = cmd.Process.Kill()
+		_ = cmd.Wait()
+		return errors.ErrEgressNotFound
+	}
 }
 
-// Gather implements the prometheus.Gatherer interface on server-side to allow aggregation of handler metrics
+func (pm *ProcessManager) GetContext(egressID string) context.Context {
```
biglittlebigben, line 93: Is this currently used?

biglittlebigben (approved):

pkg/service/process.go
biglittlebigben: It is in cloud

frostbyte73: yes, by cloud egress

PR #677, opened 2024-05-16
log keyframe interval

frostbyte73:


pkg/pipeline/builder/video.go
```diff
@@ -541,7 +541,11 @@ func (b *VideoBin) addEncoder() error {
 		}
 		x264Enc.SetArg("speed-preset", "veryfast")
 		if b.conf.KeyFrameInterval != 0 {
-			if err = x264Enc.SetProperty("key-int-max", uint(b.conf.KeyFrameInterval*float64(b.conf.Framerate))); err != nil {
+			keyframeInterval := uint(b.conf.KeyFrameInterval * float64(b.conf.Framerate))
+			if b.conf.Debug.LogKeyFrames {
```
biglittlebigben, line 545: Any reason not to always log it?

biglittlebigben (approved):

pkg/pipeline/builder/video.go
frostbyte73: trying to generally reduce noise

PR #688, opened 2024-06-03
Allow SRT output on egress server

yaruno:
TLDR; PoC of SRT support for egress server. Could use a bit of help on testing this and in general getting the test framework running.

Gstreamr seems to come with SRT support and adding it to the livekit egress server would be a great addition alongside the existing output options. Without knowing Golang too deeply I've appended the following pieces that 'should' allow usage of SRT as an egress option. However I'm bit struggling on getting this tested out. Any pointers on how to get this running and tested?


pkg/config/output.go
```diff
@@ -84,7 +86,12 @@ func (p *PipelineConfig) updateEncodedOutputs(req egress.EncodedOutput) error {
 		return errors.ErrInvalidInput("multiple stream outputs")
 	}
 	if stream != nil {
-		conf, err := p.getStreamConfig(types.OutputTypeRTMP, stream.Urls)
+		outputType := types.OutputTypeRTMP
+		if len(stream.Urls) > 0 && strings.HasPrefix(stream.Urls[0], "srt://") {
```
biglittlebigben, line 90: I may be cleaner to have a `switch` statement on the prefix rather than defaulting to rtmp.

We should also validate that all the streams use the same protocol as the code currently doesn't allow mixing SRT and RTMP.

pkg/pipeline/controller.go
```diff
@@ -263,8 +264,13 @@ func (c *Controller) UpdateStream(ctx context.Context, req *livekit.UpdateStream
 
 	// add stream outputs first
 	for _, rawUrl := range req.AddOutputUrls {
+		outputType := types.OutputTypeRTMP
```
biglittlebigben, line 267: A `switch` statement may work better here as well. We also need to validate that the added URLs use the same protocol as the ones that are already setup.

pkg/pipeline/controller.go
```diff
@@ -297,7 +303,12 @@ func (c *Controller) UpdateStream(ctx context.Context, req *livekit.UpdateStream
 
 	// remove stream outputs
 	for _, rawUrl := range req.RemoveOutputUrls {
-		url, _, err := c.ValidateUrl(rawUrl, types.OutputTypeRTMP)
+		outputType := types.OutputTypeRTMP
+		if strings.HasPrefix(rawUrl, "srt://") {
```
biglittlebigben, line 307: A `switch` statement may work better here too.

pkg/pipeline/controller.go
```diff
@@ -297,7 +303,12 @@ func (c *Controller) UpdateStream(ctx context.Context, req *livekit.UpdateStream
...
+			outputType = types.OutputTypeSRT
+		}
+		url, _, err := c.ValidateUrl(rawUrl, outputType)
+		//url, _, err := c.ValidateUrl(rawUrl, types.OutputTypeRTMP)
```
biglittlebigben, line 311: Please remove the commented code

biglittlebigben (commented):
Thanks for getting this together!

In terms of testing, getting the integration test running a SRT would be a requirement to get this merged. The [mediamtx project](https://github.com/bluenviron/mediamtx) we are using to test RTMP support SRT as well, so it should be possible to do a similar test for SRT.

A few other small comments below.

pkg/config/output.go
yaruno: Changed to switch statement. Not sure how to validate that the streams use the same protocol. Would be nice if there were some unit tests to catch this edge case.

pkg/pipeline/controller.go
yaruno: same as above

pkg/pipeline/controller.go
yaruno: schanged to switch case

pkg/pipeline/controller.go
yaruno: removed

pkg/config/output.go
```diff
@@ -84,7 +86,22 @@ func (p *PipelineConfig) updateEncodedOutputs(req egress.EncodedOutput) error {
...
+		u, err := url.Parse(stream.Urls[0])
+		if err != nil {
+			return errors.ErrInvalidInput("malformed url")
+		}
+
+		var outputType types.OutputType
+		switch u.Scheme {
+		case "srt":
+			outputType = types.OutputTypeSRT
+		case "rtmp":
```
biglittlebigben, line 98: We also need to test for "rtmps"

pkg/pipeline/controller.go
```diff
@@ -263,8 +264,23 @@ func (c *Controller) UpdateStream(ctx context.Context, req *livekit.UpdateStream
...
+		u, err := url.Parse(rawUrl)
+		if err != nil {
+			return errors.ErrInvalidInput("malformed url")
+		}
+
+		var outputType types.OutputType
+		switch u.Scheme {
+		case "srt":
+			outputType = types.OutputTypeSRT
+		case "rtmp":
```
biglittlebigben, line 277: We also need to support rtmps

pkg/pipeline/controller.go
```diff
@@ -297,7 +313,23 @@ func (c *Controller) UpdateStream(ctx context.Context, req *livekit.UpdateStream
...
```
biglittlebigben, line 325: We also need to support rtmps

pkg/config/output.go
biglittlebigben: Up to this change, all streams were always RTMP, so there was no reason to check that they were all the same protocol. It is needed now that we support more than 1 protocol. Adding a unit test for this is a good idea.

pkg/config/output.go
yaruno: Would that default to same output type as rtmp or should there be a new output type for rtmps ?

biglittlebigben: All current output URL schemes: rtmp://, rtmps:// and mux:// should use OutputTypeRTMP.

PR #726, opened 2024-07-18
Adjust twitch ingest urls

frostbyte73:
Uses the ingest endpoint api to choose the best url for twitch streams


pkg/config/urls.go
```diff
@@ -0,0 +1,124 @@
...
+	"strings"
+
+	"github.com/go-jose/go-jose/v3/json"
+
+	"github.com/livekit/egress/pkg/errors"
+	"github.com/livekit/egress/pkg/types"
+	"github.com/livekit/protocol/utils"
+)
+
+var twitchEndpoint = regexp.MustCompile("^rtmps?://.*\\.contribute\\.live-video\\.net/app/(.*)( live=1)?$")
```
dennwc, line 31: A nit, but if you use backquote you won't need to escape the slash.

pkg/config/urls.go
```diff
@@ -0,0 +1,124 @@
...
+func ValidateUrl(rawUrl string, outputType types.OutputType) (string, string, error) {
+	parsed, err := url.Parse(rawUrl)
+	if err != nil {
+		return "", "", errors.ErrInvalidUrl(rawUrl, err.Error())
+	}
+
+	switch outputType {
+	case types.OutputTypeRTMP:
+		if parsed.Scheme == "mux" {
+			rawUrl = fmt.Sprintf("rtmps://global-live.mux.com:443/app/%s", parsed.Host)
```
dennwc, line 42: The `fmt.Sprintf` could be swapped to `+`. Up to you, though.

pkg/config/urls.go
```diff
@@ -0,0 +1,124 @@
...
+		if match := twitchEndpoint.FindStringSubmatch(u); len(match) > 0 && match[1] == twitchKey {
+			return u, nil
+		}
+	}
+
+	return "", errors.ErrStreamNotFound(rawUrl)
+}
+
+func updateTwitchURL(key string) (string, error) {
+	resp, err := http.Get("https://ingest.twitch.tv/ingests")
```
dennwc, line 100: Maybe we could pass context here from the `UpdateStream`? I see that it's passed to other methods of `ipcServiceClient`, so it will get cancelled there eventually. We'll just wait a bit longer here.

pkg/pipeline/controller.go
```diff
@@ -264,7 +264,7 @@ func (c *Controller) UpdateStream(ctx context.Context, req *livekit.UpdateStream
 	// add stream outputs first
 	for _, rawUrl := range req.AddOutputUrls {
 		// validate and redact url
-		url, redacted, err := c.ValidateUrl(rawUrl, types.OutputTypeRTMP)
+		url, redacted, err := config.ValidateUrl(rawUrl, types.OutputTypeRTMP)
```
dennwc, line 267: I have a question about this one. It's called for each output URL and it may eventually call `updateTwitchURL`. How likely is that its called twice? I guess it's not typical to output to multiple Twitch streams, right?

dennwc (approved):

pkg/pipeline/controller.go
frostbyte73: I've never seen anyone do it. I considered caching the results but it's a pretty cheap call and didn't seem worth the added complexity

PR #740, opened 2024-07-30
close src once recording has stopped

frostbyte73:
This way the recording status on the room doesn't lag behind while waiting for files to upload


lukasIO (approved):
nice, lgtm!

PR #754, opened 2024-08-08
Clean up stream logic

frostbyte73:


pkg/pipeline/controller.go
```diff
@@ -264,128 +266,111 @@ func (c *Controller) UpdateStream(ctx context.Context, req *livekit.UpdateStream
...
 			errs.AppendErr(err)
-		} else {
-			sendUpdate = true
 		}
 	}
 
 	if sendUpdate {
 		c.Info.UpdatedAt = time.Now().UnixNano()
-		_, _ = c.ipcServiceClient.HandlerUpdate(ctx, (*livekit.EgressInfo)(c.Info))
+		c.streamUpdates(func() {
```
biglittlebigben, line 316: All updated will be delayed by at least 500ms. Could this cause temporary odd state mismatches when accessing the stream state after updating the egress?

biglittlebigben (approved):

pkg/pipeline/controller.go
frostbyte73: it should grab the info and update time at the time it's executed, so even if that state isn't the same as when this was queued up it should always be accurate/up to date

PR #762, opened 2024-08-26
skip aws logging on successful upload

frostbyte73:


boks1971 (approved):
Is that slice going to become too large? Should we put a limit on number of messages?

boks1971 (commented):
Also, I guess all of them will have the same timestamp in the logs. Is that okay?

frostbyte73:
> Is that slice going to become too large? Should we put a limit on number of messages?

That is probably a good idea. I have no idea how many logs they put out but it seems to be a lot

pkg/pipeline/sink/uploader/s3.go
```diff
@@ -40,18 +41,48 @@ const (
...
+// S3Logger only logs aws messages on upload failure
+type S3Logger struct {
+	mu   sync.Mutex
+	msgs []string
+}
+
+func (l *S3Logger) Log(args ...interface{}) {
+	msg := "aws sdk:"
+	for range len(args) {
+		msg += " %v"
```
biglittlebigben, line 64: Tin my experience, building log strings this way cause enough men copies that it can have a significant performance impact. strings.Builder should be more efficient

pkg/pipeline/sink/uploader/s3.go
```diff
@@ -40,18 +41,48 @@ const (
...
+	}
+
+	l.mu.Lock()
+	if len(l.msgs) >= 10 {
+		l.msgs = append(l.msgs[1:], msg)
```
biglittlebigben, line 69: If my understanding of go slices is correct, this will cause the array backing to slice to grow indefinitely until the logger gets destroyed as the beginning of the array will never be reclaimed. We may want to either copy the leftover part of the slice to the beginning of the array, or use ring buffer implementation?

PR #779, opened 2024-09-19
playlist upload backoff

frostbyte73:
Uploads once every N updates where N is the number of hours (when using the default duration) that have passed


pkg/pipeline/sink/segments.go
```diff
@@ -200,6 +206,13 @@ func (s *SegmentSink) handlePlaylistUpdates(update SegmentUpdate) error {
 	return nil
 }
 
+// Each segment adds about 100 bytes in the playlist, and long playlists can get very large.
+func (s *SegmentSink) shouldUploadPlaylist() bool {
+	// with the default of 6s segments, this will upload every update for the first hour, then every
+	// other update for the second, every third update for the third, etc.
+	return s.segmentCount < 600 || s.segmentCount%(s.segmentCount/600) == 0
```
biglittlebigben, line 213: Should we cap the interval between uploads? Uploading every 30 segments for broadcasts > 30h means you can miss up to 3min of broadcast if the pod crashes. Maybe upload at least once every 10 segments?

biglittlebigben (approved):

pkg/pipeline/sink/segments.go
frostbyte73: it is capped to at least once per minute

PR #789, opened 2024-10-07
check track prefix

frostbyte73:


pkg/config/pipeline.go
```diff
@@ -332,11 +332,17 @@ func (p *PipelineConfig) Update(request *rpc.StartEgressRequest) error {
...
+				return errors.ErrInvalidInput("audio_track_id")
+			}
 			p.AudioEnabled = true
 			p.AudioTrackID = audioTrackID
 			p.AudioTranscoding = true
 		}
 		if videoTrackID := req.TrackComposite.VideoTrackId; videoTrackID != "" {
+			if strings.HasPrefix(videoTrackID, "TR_A") {
+				return errors.ErrInvalidInput("video_track_id")
+			}
```
boks1971, line 345: If track source is not set, I think the A/V prefix will not be there IIRC.

biglittlebigben: I think that's fine provided we never have a valid video track with a "A" prefix? We may not catch every error case though indeed.

biglittlebigben (approved):

pkg/config/pipeline.go
frostbyte73: I had no idea you could have a track ID without a prefix. I'll have to check it at another point then

biglittlebigben (approved):

davidzhao (approved):

PR #792, opened 2024-10-15
backup storage options

frostbyte73:


biglittlebigben (approved):
LGTM. Should we send a webhook with the filename for each file that was sent to the backup destination?

frostbyte73:
> LGTM. Should we send a webhook with the filename for each file that was sent to the backup destination?

We'll have to do something with cloud-egress/cloud-io and generate presigned download urls

frostbyte73:
Still need to figure out what to do about situations like some segments or images going to backup while the rest go to primary

PR #806, opened 2024-11-12
Avoid recording room composite before video is decoded

davidzhao:
It's likely to receive delta frames that are not decodable in the start. This would produce an undesirable black screen in the beginning of the recording.

### DIFF
--- a/template-default/src/Room.tsx
+++ b/template-default/src/Room.tsx
@@ -28,6 +28,8 @@ import { ReactElement, useEffect, useState } from 'react';
 import SingleSpeakerLayout from './SingleSpeakerLayout';
 import SpeakerLayout from './SpeakerLayout';
 
+const FRAME_DECODE_TIMEOUT = 5000;
+
 interface RoomPageProps {
   url: string;
   token: string;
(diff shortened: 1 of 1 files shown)


biglittlebigben:
This means we'll potentially cut some audio from the start of the recording, until a video becomes decodable. Is this always desirable?

I can also see how a few customers may have a broken setup where they publish a video track but never send any video so we'd want to advertise this change pretty explicitly in the changelog.

davidzhao:
It's a tricky one for sure. these are the use cases:
1. video users do not want to record a black screen as the initial thumbnail
2. audio-only users should work immediately
3. edge cases like having a video published, but no frames coming through

before this PR, it doesn't satisfy 1, which is where users have been complaining about.

For 2, we should move this to mixing within the process, instead of using templates for them.

3. may still be an issue, and I think it's handled by adding a timeout (one could argue 10s is too long, I'd be happy to lower it to 5s). I think if folks have custom use cases, they should host their own template. It feels kind of broken that the primary use case (recording a video) doesn't work correctly.

frostbyte73 (approved):

PR #817, opened 2024-11-26
Use the SDK to handle audio only room composite requests with the default template

biglittlebigben:


dennwc (approved):
Not very familiar with the codebase, but it looks good so far :+1:

pkg/config/base.go
```diff
@@ -36,12 +36,13 @@ type BaseConfig struct {
...
-	MaxUploadQueue       int            `yaml:"max_upload_queue"`       // maximum upload queue size, in minutes
-	DisallowLocalStorage bool           `yaml:"disallow_local_storage"` // require an upload config for all requests
+	Logging                      *logger.Config `yaml:"logging"`                          // logging config
+	TemplateBase                 string         `yaml:"template_base"`                    // custom template base url
+	ClusterID                    string         `yaml:"cluster_id"`                       // cluster this instance belongs to
+	EnableChromeSandbox          bool           `yaml:"enable_chrome_sandbox"`            // enable Chrome sandbox, requires extra docker configuration
+	MaxUploadQueue               int            `yaml:"max_upload_queue"`                 // maximum upload queue size, in minutes
+	DisallowLocalStorage         bool           `yaml:"disallow_local_storage"`           // require an upload config for all requests
+	EnableRoomCompositeSDKSource bool           `yaml:"enable_room_composite_sdk_source"` // attempt to render supported audio only room composite use cases using the SDK source instead of Chrome
```
frostbyte73, line 45: Would prefer this to be automatic, egress config is already pretty extensive

frostbyte73 (approved):

pkg/config/base.go
biglittlebigben: Considering there is a not insignificant risk with rolling this out, I think we need a way to enable it on a per cluster basis and to disable it easily. Is there a better way to achieve this?

frostbyte73: that's fair, can always remove it in the future

davidzhao (commented):
lg

one thing I didn't see (possibly I've missed it), how do we handle tracks that are published/unpublished after the pipeline had started? 

does `buildSDKInput` get called again to cycle through all audio tracks again?

biglittlebigben:
It ends up using the same logic as participant egress to handle track updates. I don't quite recall the flow at this time.

PR #850, opened 2025-01-30
Chrome retries

frostbyte73:
* Adds retries on page load timeouts
* Starts the timer after chrome has been initialized

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -92,7 +92,7 @@ func ErrProcessFailed(process string, err error) error {
 	return psrpc.NewErrorf(psrpc.Internal, "failed to launch %s: %v", process, err)
 }
 
-func ErrChromeFailedToStart(err error) error {
+func ChromeError(err error) error {
 	return psrpc.NewError(psrpc.Internal, err)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/errors/errors.go
```diff
@@ -109,8 +109,15 @@ var (
 	ErrShuttingDown               = psrpc.NewErrorf(psrpc.Unavailable, "server is shutting down")
 )
 
-func ErrPageLoadFailed(err string) error {
-	return psrpc.NewErrorf(psrpc.InvalidArgument, "template page load failed: %s", err)
+func PageLoadError(err string) error {
+	if strings.HasPrefix(err, "page load error ") {
```
dennwc, line 113: nit: can be replaced with `CutPrefix`

pkg/pipeline/source/web.go
```diff
@@ -262,81 +255,99 @@ func (s *WebSource) launchChrome(ctx context.Context, p *config.PipelineConfig,
...
-				case startRecordingLog:
-					logger.Infow("chrome: START_RECORDING")
-					if s.startRecording != nil {
-						select {
-						case <-s.startRecording:
-							continue
-						default:
-							close(s.startRecording)
-						}
+		chromeCtx, chromeCancel := chromedp.NewContext(allocCtx)
```
dennwc, line 273: This is a pretty large code block inside a loop. Maybe deserves a separate function?

I cannot quickly tell if `chromeCancel` is always called right now, for example.

dennwc (approved):

biglittlebigben (approved):

PR #863, opened 2025-02-21
Update server.go

xtqxk:
TemplateBase not used.


biglittlebigben (commented):
TemplateBase is meant as a way to point egress to a template location potentially hosted on a different server (vs the current pod). What issue are you trying to solve? Binding the server on a different interface than localhost also has security implication.

PR #883, opened 2025-03-20
Fix sdk room composite

frostbyte73:


pkg/pipeline/source/sdk.go
```diff
@@ -314,6 +296,29 @@ func (s *SDKSource) awaitParticipantTracks(identity string) (uint32, uint32, err
...
+		case sub := <-s.errors:
+			if sub.err != nil {
+				return sub.err
+			}
+			subscribed++
+			if subscribed == expected {
+				return nil
+			}
+		case <-deadline:
+			return nil
```
boks1971, line 317: no error return here is okay?

pkg/pipeline/source/sdk.go
```diff
@@ -217,37 +217,37 @@ func (s *SDKSource) joinRoom() error {
...
-	trackCount := len(tracks)
-loop:
-	for i := 0; i < trackCount; i++ {
+	// lock any incoming subscriptions
+	s.subLock.Lock()
+	defer s.subLock.Unlock()
+
+	for {
 		select {
+		// check errors from any tracks published in the meantime
```
boks1971, line 240: maybe rename variable to `pub`?

boks1971 (approved):

pkg/pipeline/source/sdk.go
boks1971: ah got it

pkg/pipeline/source/sdk.go
boks1971: guess these are remote pubs and sub here. Got confused by that check. Did not exactly understand what errors this is catching.

frostbyte73: probably makes sense to rename `s.errors`

PR #885, opened 2025-03-26
Do not log an empty string an error for glib errors/warnings

biglittlebigben:
glib prints 2 carriage returns.


pkg/logging/handler.go
```diff
@@ -27,7 +27,11 @@ var actions = map[string]int{
 func NewHandlerLogger(handlerID, egressID string) *medialogutils.CmdLogger {
 	l := logger.GetLogger().WithValues("handlerID", handlerID, "egressID", egressID)
 	return medialogutils.NewCmdLogger(func(s string) {
-		lines := strings.Split(strings.TrimSuffix(s, "\n"), "\n")
+		// glib inserts 2 carriage returns at the end of its warning/error logs
+		for strings.HasSuffix(s, "\n") {
```
dennwc, line 31: Cut be a single call to `strings.TrimRight` or `strings.CutPrefix` in a loop.

dennwc (approved):

PR #892, opened 2025-04-10
Fix race between track subscriptions and audio bin creation

frostbyte73:


pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -150,6 +151,15 @@ func (w *AppWriter) start() {
 		go w.logStats()
 	}
 
+	go func() {
+		<-w.callbacks.BuildReady
```
dennwc, line 155: Is there any other signal we can select on? `endStream` or `finished` maybe?

pkg/pipeline/source/sdk.go
```diff
@@ -639,8 +639,9 @@ func (s *SDKSource) onTrackFinished(trackID string) {
 		writer.Drain(true)
 		active := s.active.Dec()
 		if s.RequestType == types.RequestTypeParticipant || s.RequestType == types.RequestTypeRoomComposite {
-			s.callbacks.OnTrackRemoved(trackID)
 			s.sync.RemoveTrack(trackID)
+			<-s.callbacks.BuildReady
```
dennwc, line 643: Same here, any other channel we can select on?

dennwc (approved):

pkg/pipeline/source/sdk.go
frostbyte73: not for this one, this needs to happen even if `closed` was already broken

PR #895, opened 2025-04-18
Add support for IMAGE_SUFFIX_NONE_OVERWRITE

biglittlebigben:
Allow only keeping the latest image with ImageOutput


frostbyte73 (changes requested):
DeleteAfterUpload should probably be false when using override

biglittlebigben:
> DeleteAfterUpload should probably be false when using override

Can you elaborate on this? Why would be want to keep the original image locally in this mode? The use case we are trying to support is somebody trying to have a thumbnail of the stream updated every x seconds, always at the same origin location, they can point their CDN to.

frostbyte73:
> > DeleteAfterUpload should probably be false when using override
> 
> 
> 
> Can you elaborate on this? Why would be want to keep the original image locally in this mode? The use case we are trying to support is somebody trying to have a thumbnail of the stream updated every x seconds, always at the same origin location, they can point their CDN to.

If an upload takes too long, and the new image is created before the previous upload completes and deletes the file, it can delete the newly generated one. It might also break if trying to upload the previous one while locally overwriting it

biglittlebigben:
> > > DeleteAfterUpload should probably be false when using override
> > 
> > 
> > Can you elaborate on this? Why would be want to keep the original image locally in this mode? The use case we are trying to support is somebody trying to have a thumbnail of the stream updated every x seconds, always at the same origin location, they can point their CDN to.
> 
> If an upload takes too long, and the new image is created before the previous upload completes and deletes the file, it can delete the newly generated one. It might also break if trying to upload the previous one while locally overwriting it

I believe the filenames locally are still unique (sequence number suffix: https://github.com/livekit/egress/blob/284051058788e6c8958008f13b2c96232e8fa7f0/pkg/pipeline/builder/image.go#L128). Since there is no upload queue and all uploads happen in parallel, there is indeed a small chance of an image being overwritten by an older one. Renaming the file before the upload to a common filename may not help because the uploads are likely to open the file at the start of the upload, and hang onto the file descriptor. Replacing a file doesn't change the content of already open files.

I believe the best fix would be to have make upload descriptor a singleton in the uploader, specifically in this mode. The uploader only uploads at most one file. Another upload can be queued, but if there is already a pending upload, the queued one gets replaced.

If we believe this improvement is needed before reloading this feature, I can implement the above now. Otherwise I can commit the current implementation and create a ticket.

biglittlebigben:
Never mind, I while looking into implementing the above, I realized that we already never upload more than 1 image at the same time (createdImages channel). Not uploading outdated images in this mode would still be an improvement, but at least I don't believe there is a race.

frostbyte73 (approved):

PR #901, opened 2025-05-12
update all deps and go version

matkam:


pkg/pipeline/source/web.go
```diff
@@ -302,7 +302,7 @@ func (s *WebSource) navigate(chromeCtx context.Context, chromeCancel context.Can
 		switch ev := ev.(type) {
 		case *runtime.EventConsoleAPICalled:
 			if s.chromeLog != nil {
-				if b, err := ev.MarshalJSON(); err == nil {
+				if b, err := json.Marshal(ev); err == nil {
```
matkam, line 305: @frostbyte73 @biglittlebigben is this ok?

biglittlebigben: Should be. We won't know until we test though.

biglittlebigben (approved):

PR #910, opened 2025-05-21
SDK timing

frostbyte73:
also cleaned up some constants


pkg/pipeline/builder/video.go
```diff
@@ -259,17 +259,6 @@ func (b *VideoBin) buildSDKInput() error {
 
 	if b.conf.VideoDecoding {
 		b.bin.SetGetSrcPad(b.getSrcPad)
```
biglittlebigben, line 261: Is this special handling not needed anymore?

biglittlebigben (approved):

pkg/pipeline/builder/video.go
frostbyte73: I'm not sure it ever was, we can just EOS the videoTestSrc bin as normal and skip the EOS on all the appSrcBins

PR #924, opened 2025-06-16
fix audio mixer

frostbyte73:


pkg/pipeline/builder/audio.go
```diff
@@ -278,6 +278,9 @@ func (b *AudioBin) addMixer() error {
 	if err = audioMixer.SetProperty("latency", uint64(b.conf.Latency.AudioMixerLatency)); err != nil {
 		return errors.ErrGstPipelineError(err)
 	}
+	if err = audioMixer.SetProperty("alignment-threshold", uint64(b.conf.Latency.AudioMixerLatency)); err != nil {
```
biglittlebigben, line 281: Is this the magic property? It doesn't seem documented. What does it do?

biglittlebigben (approved):

pkg/pipeline/builder/audio.go
frostbyte73: It is. It's set to 40ms by default, which means when two buffers arrive from different tracks and their timestamps are more than 40ms apart, it drops and/or adds gaps to try to align them

PR #936, opened 2025-06-25
memory updates

frostbyte73:


pkg/stats/monitor.go
```diff
@@ -418,7 +424,7 @@ func (m *Monitor) GetAvailableMemory() float64 {
 	defer m.mu.Unlock()
 
 	if m.cpuCostConfig.MaxMemory == 0 {
-		return math.MaxFloat64
+		return float64(memory.FreeMemory()) / gb
```
biglittlebigben, line 427: Does is FreeMemory cgroup aware? Does it return the total amount of memory on the system/assigned to the cgroup, or the amount of memory allocatable in the system?

frostbyte73: It's not cgroup aware, but afaik we don't have anything for that, otherwise it would replace MaxMemory in the cost config

biglittlebigben (approved):

PR #961, opened 2025-08-11
set status to failed when an egress is killed

frostbyte73:


pkg/handler/handler_ipc.go
```diff
@@ -106,3 +109,15 @@ func renderMetrics(metrics []*dto.MetricFamily) (string, error) {
...
 	return writer.String(), nil
 }
+
+func (h *Handler) KillEgress(ctx context.Context, req *ipc.KillEgressRequest) (*emptypb.Empty, error) {
+	ctx, span := tracer.Start(ctx, "Handler.KillEgress")
+	defer span.End()
+
+	<-h.initialized.Watch()
+
+	h.controller.SendEOS(ctx, livekit.EndReasonKilled)
```
milos-lk, line 119: just for my understanding -  does this drain the pipeline before returning - i.e do we completely stop the pipeline before returning? what would happen if we don't send EOS?

frostbyte73: yes, this will still stop the pipeline normally and upload the file(s)

milos-lk: I see - this will cause pipeline to eventually wind down asynchronously.
the only possibility for this not to work is that EOS event isn't propagated through - if the pipeline stalls for some reason - not sure if we observed that before (in which case a small timeout in the kill function - with the hard kill on expiration could also be considered)

frostbyte73: SendEOS will kill the pipeline after 30s, I think it's better to give it a chance to upload since there generally isn't anything wrong with the recording with these

PR #967, opened 2025-08-16
Bump server-sdk-go to 2.10.0

milos-lk:
Reducing max allowed PTS diff tolerance to 1s by using new synchronizer constructor. Higher values could cause gstreamer pipeline stalls on unexpected PTS jumps.
New jitter buffer API requires callback instead of a channel for getting samples and since samples could be dropped in the callback implementation - adding writer stats to keep track of packets received from the buffer but dropped before pushing to the app source.


pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -76,6 +76,11 @@ type AppWriter struct {
 	draining     core.Fuse
 	endStream    core.Fuse
 	finished     core.Fuse
+	stats        appWriterStats
+}
+
+type appWriterStats struct {
```
biglittlebigben, line 82: Since this is modified from a callback where we do not control the calling goroutine, this should probably be an atomic (or we should use a lock)

milos-lk: updated

PR #974, opened 2025-08-24
Handle small audio gaps by the signal time stretching/compression instead of having discontinuities

milos-lk:
Using pitch gstreamer element for controlling audio signal tempo that is then driven by a drift detected by RTCP sender reports.
On SRs drifts, app source will continue producing buffers with monotonically increasing PTS. The reported drift will be handled either by slowing down the audio signal (if it's positive) or speeding it up. The pitch element will relatively gently change the rate (by 10%) without affecting pitch.
Pitch elements require F32 on its sink pads so adding caps filters around it to convert from and to S16.
Due to max pipeline latency - accumulated drift (if negative) could only raise up to that value after which we walk off the audio mixer latency cliff - so adding logic to guard against big cumulative drifts. An option for future improvements would be dynamically increasing pipeline latency and buffer sizes in that case but that could also go up to a certain point.

An example of a pipeline:
File_1_ParticipantComposite_H264

### DIFF
--- a/pkg/pipeline/builder/audio.go
+++ b/pkg/pipeline/builder/audio.go
@@ -37,6 +37,10 @@ const (
 
 	leakyQueue    = true
 	blockingQueue = false
+
+	speedUpTempo       = 1.1
+	slowDownTempo      = 0.9
+	compensationFactor = 10
 )
 
 type AudioBin struct {
@@ -66,16 +70,20 @@ func (a *audioPacer) start(drift time.Duration) {
 	if a.pitch == nil || drift == 0 {
 		return
 	}
-
-	logger.Debugw("Handling audio time drift", "drift", drift)
-	rate := 1.1
-	if drift > 0 {
-		rate = 0.9
+	if a.active {
+		logger.Errorw(
+			"starting audio pacer, but it's already active",
+			errors.New("tempo controller bug"),
+		)
+		return
 	}
-	a.remaining = 10 * drift.Abs()
 
+	rate := speedUpTempo
+	if drift > 0 {
+		rate = slowDownTempo
+	}
+	a.remaining = compensationFactor * drift.Abs()
 	a.pitch.SetArg("tempo", fmt.Sprintf("%.2f", rate))
-
 	a.active = true
 
 }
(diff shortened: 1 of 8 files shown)


pkg/pipeline/builder/audio.go
```diff
@@ -36,6 +37,10 @@ const (
 
 	leakyQueue    = true
 	blockingQueue = false
+
+	speedUpTempo       = 1.1
```
biglittlebigben: This should ideally be configurable. I expect 10% to be noticeable, but maybe still acceptable.

pkg/pipeline/builder/audio.go
```diff
@@ -46,6 +51,62 @@ type AudioBin struct {
...
+
+	audioPacer audioPacer
+}
+
+type driftProcessNotifier interface {
+	DriftProcessed()
+}
+
+type audioPacer struct {
+	pitch     *gst.Element
```
biglittlebigben: Does the audioPacer need to be thread safe? I believe it is accessed both from a GST probe and the RTCP report handler.

biglittlebigben (changes requested):
Thanks! Would it be possible to make enabling this configurable, to make roll out safer?

pkg/pipeline/builder/audio.go
milos-lk: I built it in a way not to require sync - multiple concurrent SRs will be serialized and taken one by one. However there is one good point here - making its active flag atomic (introducing memory barrier) so we avoid reading stale value in the pad probe forever (compiler optimizations). Will update that. Otherwise I see no other thread safety issues.

pkg/pipeline/builder/audio.go
milos-lk: updated - added the config

pkg/pipeline/builder/audio.go
```diff
@@ -214,14 +281,29 @@ func (b *AudioBin) addAudioAppSrcBin(ts *config.TrackSource) error {
 		return errors.ErrNotSupported(string(ts.MimeType))
 	}
 
-	if err := addAudioConverter(appSrcBin, b.conf, b.getChannel(ts), blockingQueue); err != nil {
+	addAudioConvertFunc := addAudioConverter
+	if b.conf.AudioTempoController.Enabled {
```
biglittlebigben, line 285: We should make sure we log whether it's enabled and the rate once per egress.

PR #989, opened 2025-09-04
Switch from react-scripts to vite for egress-templates build

biglittlebigben:
Our current templates build has several vulnerabilities in libraries imported by react-scripts. In several cases, the fixed version is not compatible with the latest version of react-scripts which was deprecated 3 years ago.

In an attempt to fix this, this switches the template framework from react-scripts to vite. Not being a frontend developer, this applies pretty directly a migration guide from the interweb, so this may not adopt best practices... Particularly, there are no test and lint targets, but I could not find any template test in the repository in the first place.

The resulting build seems to work and to fix all the reported vulnerabilities.


frostbyte73:
Looks like the template test script node version needs to be updated

biglittlebigben:
New version has is a new Vite project created from their template with the files moved over

template-default/package.json
```diff
@@ -5,53 +5,32 @@
...
-    "@livekit/components-react": "^2.9.4",
-    "@livekit/components-styles": "^1.1.5",
+    "@livekit/components-react": "^2.9.14",
+    "@livekit/components-styles": "^1.1.6",
     "@livekit/egress-sdk": "^0.2.1",
-    "livekit-client": "^2.13.0",
-    "react": "^19.1.0",
-    "react-dom": "^19.1.0",
-    "react-scripts": "5.0.1"
+    "@vitejs/plugin-react": "^5.0.2",
```
lukasIO: nit, this should not be needed here, it's a devDependency

lukasIO (approved):
lgtm apart from the build CI failing.

milos-lk (approved):

1egoman (approved):

template-default/package.json
biglittlebigben: Interesting. This was setup by the Vite project creation template...

PR #992, opened 2025-09-08
Enabling PLC for opus decoder

milos-lk:
Adding opuseparse element before the opus decoder in order to be able to determine buffer durations before the decoder and have ability to signal gaps should it results in smaller (up to 5 frames) gaps. In order to make the PLC logic effective a gap event needs to be signaled to the decoder in addition to enabling its `plc` property.
A periodic poll of opus decoder stats is also added and exposed as an element message used to update audio qos metrics.

### DIFF
--- a/pkg/pipeline/builder/audio.go
+++ b/pkg/pipeline/builder/audio.go
@@ -17,6 +17,7 @@ package builder
 import (
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/go-gst/go-gst/gst"
 
@@ -36,6 +37,8 @@ const (
 
 	leakyQueue    = true
 	blockingQueue = false
+
+	opusPlcMaxFrames = 4
 )
 
 type AudioBin struct {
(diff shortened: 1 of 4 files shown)


pkg/pipeline/builder/audio.go
```diff
@@ -397,3 +413,66 @@ func subscribeForQoS(mixer *gst.Element) {
...
+
+		pts := time.Duration(buf.PresentationTimestamp())
+		dur := time.Duration(buf.Duration())
+
+		if dur <= 0 {
+			// Fallback if TOC wasn’t parsed (shouldn’t happen with opusparse)
+			if lastDur > 0 {
+				dur = lastDur
+			} else {
+				dur = 20 * time.Millisecond
```
boks1971: :nit: maybe make a const for default opus duration?

pkg/pipeline/builder/audio_stats.go
```diff
@@ -0,0 +1,129 @@
...
+		return "", false
+	}
+}
+
+/*** minimal parser for the serialized GstStructure ***/
+
+func parseStatsString(s string) (OpusDecStats, error) {
+	var st OpusDecStats
+
+	getU64 := func(k string) (uint64, error) {
```
boks1971: are these useful as some util in a utils package?

milos-lk: interesting idea - a generic set of functions for finding an integer in a given regex might be interesting to have somewhere available - is there already a utils package you could recommend for this or you maybe referred to creating a utils package in this project?

boks1971: We put a lot of utils in `protocol` repo. Would be good to put it there.

milos-lk: I tried to extract it here: https://github.com/livekit/protocol/pull/1195

milos-lk: updated

pkg/pipeline/builder/audio.go
```diff
@@ -36,6 +37,11 @@ const (
 
 	leakyQueue    = true
 	blockingQueue = false
+
+	opusPlcMaxFrames          = 5
+	opusDecStatsPollInterval  = time.Second * 5
+	opusDecPlcMaxJitter       = 3 * time.Millisecond
```
frostbyte73, line 43: why 3ms?

milos-lk: semi arbitrary chosen value :)
The case I had in mind was - if the gap is almost one frame (e.g., 18–19 ms when the true frame is 20 ms) still count it as a loss. That probably won't happen as long as we lose full packets. Added for the cases when wall-clock PTS adjustment kicks in (even then it's going to be extremely rare).

PR #1002, opened 2025-09-14
Snap audio PTS back to the regular time grid

milos-lk:
Due to the opus DTX - cadence of  PCM samples arriving at mixer wasn't regular which could cause audio build up on on of the mixer pads and problems with syncing different tracks

Solution here snaps decoded PCM buffer timestamps back to the regular grid using audiorate element.


boks1971 (approved):

biglittlebigben:
Is it expected for the opus decoder not to emit silence between DTX packets?

PR #1004, opened 2025-09-15
Speed up build time

milos-lk:
Contains following changes
- use linkdata/deadlock - does the same thing as existing manual logic of adding/removing go-deadlock
- use go mod and go build cache
- optimizing Docker file for tests to avoid frequent cache busting
- using docker registry for build cache for CI

Major improvements are with incremental builds where the speed up is 10x (from 240s to 25s).

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -40,7 +40,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
         run: build/test/fetch-media-samples.sh
-      
+ 
       - uses: actions/cache@v4
         with:
           path: |
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -81,6 +81,10 @@ jobs:
           platforms: linux/amd64
           tags: ${{ steps.docker-md.outputs.tags }}
           labels: ${{ steps.docker-md.outputs.labels }}
+          build-args: |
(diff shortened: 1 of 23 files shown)


.github/workflows/test-integration.yaml
```diff
@@ -78,17 +78,28 @@ jobs:
...
           context: .
           file: ./build/test/Dockerfile
           push: true
           platforms: linux/amd64
           tags: ${{ steps.docker-md.outputs.tags }}
           labels: ${{ steps.docker-md.outputs.labels }}
           build-args: |
+            DEADLOCK=1
             TEMPLATE_TAG=${{ steps.template-tag.outputs.template_tag }}
+          cache-from: type=registry,ref=livekit/egress-integration:buildcache
```
biglittlebigben: I used to use the registry to cache e2e builds. We ended up with a 1.5TB docker repository. Is there a strategy to clean up old caches?

milos-lk: that's what I wanted to check here as well - thanks for bringing this up. Based on the comment I assume we don't have a periodic cleanup process in place.  I can go ahead and disable cache for CI builds for now until I add a cleanup workflow later. wdyt?

biglittlebigben: Yes, this may be safer...

milos-lk: updated

PR #1007, opened 2025-09-18
Adding MP3 to the supported audio types for audio only recordings

milos-lk:
Extending egress to support mp3 as output audio codec for file ~~and hls~~ output (audio only).
Depends on:
https://github.com/livekit/protocol/pull/1217

A sample of how pipeline looks lile:
image


pkg/pipeline/builder/file.go
```diff
@@ -55,14 +59,24 @@ func BuildFileBin(pipeline *gstreamer.Pipeline, p *config.PipelineConfig) (*gstr
 	if err = sink.SetProperty("sync", false); err != nil {
 		return nil, errors.ErrGstPipelineError(err)
 	}
+
+	if o.OutputType == types.OutputTypeMP3 {
```
biglittlebigben: Would using an `identity` plugin here allow keeping the factory code flow similar to all other formats? I've think it would make the code more maintainable in the long run.

pkg/pipeline/builder/segment.go
```diff
@@ -59,8 +74,21 @@ func BuildSegmentBin(pipeline *gstreamer.Pipeline, p *config.PipelineConfig) (*g
 	if err = sink.SetProperty("send-keyframe-requests", true); err != nil {
 		return nil, errors.ErrGstPipelineError(err)
 	}
-	if err = sink.SetProperty("muxer-factory", "mpegtsmux"); err != nil {
-		return nil, errors.ErrGstPipelineError(err)
+
+	if useMP3 {
```
biglittlebigben: Is there any pipeline step adding the PRIV ID3 tag with the first timestamp in the segment? (https://www.rfc-editor.org/rfc/rfc8216#section-3.4).

Stepping back, I'm not certain adding support for segmented mp3 files over HLS is worth the effort: Nobody asked for it, it a very legacy use case at this point and if we add it we'll need to maintain, support and document it.

pkg/config/pipeline.go
```diff
@@ -480,6 +480,14 @@ func (p *PipelineConfig) validateAndUpdateOutputParams() error {
 		}
 	}
 
+	// MP3 audio codec is compatible with HLS output if it's audio only
+	if p.VideoEnabled && p.AudioOutCodec == types.MimeTypeMP3 {
```
boks1971: should this be `!p.VideoEnabled`?

pkg/pipeline/builder/file.go
milos-lk: we would still need special case handling here:
```
b.SetGetSrcPad(func(name string) *gst.Pad {
	return mux.GetRequestPad(name + "_%u")
})
```

identity doesn't provide request pads - that's actually the first thing I wanted to write but hit a linking error. We could still put identity in the switch case above but we would need another if mp3 check inside SetGetSrcPad to return its static pad. Not sure if that's more maintainable though - wdyt?

pkg/pipeline/builder/segment.go
milos-lk: 👍  yeah - that's a good point - it adds a bit of complexity which if not needed I would also not put in now. Will go ahead and revert hls changes.
thanks for pointing out the ID3 priv tag - noted.

pkg/config/pipeline.go
milos-lk: this was meant to guard creation of the pipeline if video is enabled and and codec is mp3. But thinking about it - it doesn't need inner if. could immediately return the error. Anyways I think I will revert HLS changes (leave mp3 only supported for files) so this one will go away.

pkg/pipeline/builder/file.go
milos-lk: on the other side we could apply the fundamental theorem of software engineering - to make changes inside file sink minimal - pushed the change - let me know what you think about it.

pkg/pipeline/builder/segment.go
milos-lk: reverted HLS for now

biglittlebigben (approved):

PR #1010, opened 2025-09-23
Register all GO process metrics 

biglittlebigben:


pkg/service/metrics.go
```diff
@@ -38,6 +39,9 @@ type MetricsService struct {
 }
 
 func NewMetricsService(pm *ProcessManager) *MetricsService {
+	prometheus.Unregister(prometheus.NewGoCollector())
```
milos-lk, line 42: why do we pass a new instance of collector to the `Unregister` function? Maybe I am not reading it correctly - wouldn't this do nothing since that specific instance we just created isn't registered yet? Is it done so the call is idempotent wrt interactions with metrics?

pkg/service/metrics.go
```diff
@@ -38,6 +39,9 @@ type MetricsService struct {
...
+	prometheus.MustRegister(collectors.NewGoCollector(collectors.WithGoCollectorRuntimeMetrics(collectors.MetricsAll)))
```
milos-lk, line 43: maybe it's safer to use Register instead (and ignore already registered error)? MustRegister will cause panic in case of errors as far as I remember

pkg/service/metrics.go
biglittlebigben: This is using an established pattern to deregister Go metrics that are registered by default on launch, and reregister them along several other ones: https://github.com/livekit/cloud/blob/9694ea3ab8c23bb8e985e078911870587295a2f7/pkg/telemetry/prometheus/node.go#L18.

Is there a better way to achieve this?

pkg/service/metrics.go
biglittlebigben: We use MustRegister with more or less all our prometheus metrics. This code runs on service startup. A panic would prevent the pod from coming out of the "pending state", allowing a quick detection of a regression. Pod initialization is typically the only time when we explicitly panic.

pkg/service/metrics.go
milos-lk: just doublechecked - it matches based on desc id which would remove what we want:
https://github.com/prometheus/client_golang/blob/a974e0d45e0aa54c65492559114894314d8a2447/prometheus/registry.go#L378

haven't seen this pattern yet - my first thought was using brand new Registry but this works as well so 👍

PR #1011, opened 2025-09-23
Enable sync OnRTCP callback in any case

milos-lk:
Room composites and cases where audio tempo controller is enabled are handled WithAudioPTSAdjustmentDisabled sync option which effectively disables PTS adjustments but has a nice effect of the callback still being invoked, offset being calculated and logged.

### DIFF
--- a/pkg/pipeline/source/sdk.go
+++ b/pkg/pipeline/source/sdk.go
@@ -93,7 +93,7 @@ func NewSDKSource(ctx context.Context, p *config.PipelineConfig, callbacks *gstr
 		// to avoid gaps in the audio stream
 		opts = append(opts, synchronizer.WithAudioPTSAdjustmentDisabled())
 		if p.AudioTempoController.Enabled {
-			logger.Debugw("audio tempo controller enabled", "adjustment rate", p.AudioTempoController.AdjustmentRate)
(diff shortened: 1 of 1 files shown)


biglittlebigben (approved):

pkg/pipeline/source/sdk.go
```diff
@@ -88,11 +88,13 @@ func NewSDKSource(ctx context.Context, p *config.PipelineConfig, callbacks *gstr
...
-	if p.AudioTempoController.Enabled {
-		// perform signal time comression/steatching instead of timestamp manipulation
-		// on RTCP sender reports
-		logger.Debugw("audio tempo controller enabled", "adjustment rate", p.AudioTempoController.AdjustmentRate)
+	if p.RequestType == types.RequestTypeRoomComposite || p.AudioTempoController.Enabled {
+		// in case of room composite don't adjust audio timestamps on RTCP sender reports,
+		// to avoid gaps in the audio stream
 		opts = append(opts, synchronizer.WithAudioPTSAdjustmentDisabled())
+		if p.AudioTempoController.Enabled {
+			logger.Debugw("audio tempo controller enabled", "adjustment rate", p.AudioTempoController.AdjustmentRate)
```
boks1971: :nit: I have found it easier logging keys without spaces for filtering, something like `adjustmentRate` in this case.

boks1971 (approved):

PR #1014, opened 2025-09-25
Introduce pacer for consuming RTP packets after jitter buffer

milos-lk:
Makes sure bursty packet arrival is absorbed - preventing a track timeline to go outside of the acceptable time window
Uses:
https://github.com/livekit/media-sdk/pull/24


milos-lk:
I am going to close this PR as it won't actually fix the issue with bursts of delayed data.

PR #1015, opened 2025-09-25
Use arrival time and one way delay estimation in PTS calculation.

boks1971:
Also handle old packet and out-of-order packet errors as non-fatal, i. e. pipeline continues to run.


milos-lk:
should we also add these configs (e.g WithPreJitterBufferReceiveTimeEnabled..) to this set of changes?

boks1971:
> should we also add these configs (e.g WithPreJitterBufferReceiveTimeEnabled..) to this set of changes?

Yes, I did not know where they were. I will look for the audio pts disable example and add it there.

boks1971:
> > should we also add these configs (e.g WithPreJitterBufferReceiveTimeEnabled..) to this set of changes?
> 
> Yes, I did not know where they were. I will look for the audio pts disable example and add it there.

oh I don't see the similar structure. @milos-lk , would adding opts to the `SDKSource` be the right thing to do?

pkg/config/service.go
```diff
@@ -171,6 +172,9 @@ func (c *ServiceConfig) InitDefaults() {
 	if c.Latency.RTPMaxAllowedTsDiff == 0 {
 		c.Latency.RTPMaxAllowedTsDiff = defaultRTPMaxAllowedTsDiff
 	}
+	if c.Latency.OldPacketThreshold == 0 {
```
biglittlebigben, line 179: Documentation indicates that 0 disables drops. Is it the case here?

boks1971: There are two settings now @biglittlebigben . The current mode of operation gets two extra settings. One to enable use of receive time of packets. If that is enabled, this has an effect. If not, this does not.

For the new mode using one-way-delay estimation, it always uses arrival time of packet. Wanted to enable this by default in the new mode. So, setting default here if not set. The old mode won't be affected unless arrival time is used.

More details in this PR - https://github.com/livekit/server-sdk-go/pull/741

PR #1024, opened 2025-10-03
Drop old samples when sample queue overflows.

boks1971:
Replace channel with a simple linked list that allows dropping older packets when the queue overflows.


pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -281,18 +296,39 @@ func (w *AppWriter) handleReadError(err error) {
...
 	}
-
+	// drop old samples if queue is overflowing
+	for w.samplesLen > cSamplesQueueDepth {
+		if w.samplesHead != nil {
+			w.samplesHead = w.samplesHead.next
+			w.samplesLen--
+			w.logger.Warnw("buffer full, dropping sample", nil)
+		}
+		if w.samplesHead == nil {
```
paulwe, line 327: is this condition possible while `samplesLen > 0`?

paulwe (approved):
no tests?

boks1971:
> no tests?

have to write 😬

pkg/pipeline/source/sdk/appwriter.go
boks1971: not possible, just being paranoid :-) like if `cSamplesQueueDepth` is set to 0 kind of paranoid.

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -281,18 +296,40 @@ func (w *AppWriter) handleReadError(err error) {
...
+		w.samplesTail = item
+		w.samplesLen++
 	}
-
+	// drop old samples if queue is overflowing
+	for w.samplesLen > cSamplesQueueDepth {
+		if w.samplesHead != nil {
+			w.samplesHead = w.samplesHead.next
+			w.samplesLen--
+			w.stats.packetsDropped.Add(uint64(len(sample)))
```
milos-lk: I think we just increment this with the len of discarded samples - right?

boks1971: yep, my bad, cut and paste 🤦 . Will fix.

boks1971: Fixing in https://github.com/livekit/egress/pull/1024/commits/a3082ed2ef055571ba6bb24688e95798a3d48ba0

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -303,21 +341,42 @@ func (w *AppWriter) pushSamples() {
...
 	}
 
-	end := w.endStream.Watch()
 	for {
-		select {
-		case <-end:
-			return
-		case sample := <-w.samples:
-			for _, pkt := range sample {
+		w.samplesLock.Lock()
```
milos-lk, line 345: not sure if this is going to make it a bit simpler:
```
for {
    w.samplesLock.Lock()
    for w.samplesHead == nil && !w.endStream.IsBroken() {
        w.samplesCond.Wait()
    }
    if w.endStream.IsBroken() {
        w.samplesLock.Unlock()
        return
    }

    item := w.samplesHead
    w.samplesHead = item.next
    w.samplesLen--
    if w.samplesHead == nil {
        w.samplesTail = nil
    }
    w.samplesLock.Unlock()

    for _, pkt := range item.sample {
        if err := w.pushPacket(pkt); err != nil {
            if !utils.ErrorIsOneOf(err, synchronizer.ErrPacketOutOfOrder, synchronizer.ErrPacketTooOld) {
                w.draining.Break()
                w.endStream.Break()
                // wake any waiter
                w.samplesLock.Lock()
                w.samplesCond.Broadcast()
                w.samplesLock.Unlock()
                break
            }
        }
    }
}
```

milos-lk (approved):

pkg/pipeline/source/sdk/appwriter.go
boks1971: Looks good. Will replace.

boks1971: Added in this commit - https://github.com/livekit/egress/pull/1024/commits/abb7067d7ae73cff136da09cebc631e0110c6d22

PR #1025, opened 2025-10-03
Adding gstreamer pipeline running/playhead time accessors

milos-lk:
Using gst time to check against packets pushed to the pipeline - could flag the problems with lagging sources more specifically than mixer qos metrics (which would count muting interval as dropped).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/livekit/media-sdk v0.0.0-20251006100658-7bffd4440294
 	github.com/livekit/protocol v1.42.2-0.20251016024155-8cf58ff15ac6
 	github.com/livekit/psrpc v0.7.0
-	github.com/livekit/server-sdk-go/v2 v2.11.4-0.20251020194717-b8a10f2b31ad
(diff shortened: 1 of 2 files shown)


pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -320,7 +363,6 @@ func (w *AppWriter) onPacket(sample []jitter.ExtPacket) {
 		if w.samplesHead != nil {
 			itemToDrop := w.samplesHead
 			w.samplesHead = w.samplesHead.next
-			w.samplesLen--
```
boks1971, line 339: Is this change and the one above okay?

boks1971 (approved):

pkg/pipeline/source/sdk/appwriter.go
milos-lk: oh - no! my bad - merge conflict resolution messed this up - fixing, thanks for spotting this

milos-lk: updated

PR #1026, opened 2025-10-04
Adding burst estimator

milos-lk:
Using burst estimator to avoid track sync issues if a track is subscribed to at the time of packet burst due to retransmission.

https://github.com/livekit/server-sdk-go/pull/759


pkg/pipeline/source/sdk/burst_estimator.go
```diff
@@ -0,0 +1,110 @@
...
+		b.lastArrival = receivedAt
+		b.hasLast = true
+		return false
+	}
+
+	tsDelta := pkt.Timestamp - b.lastTS
+	arrivalDelta := receivedAt.Sub(b.lastArrival)
+
+	b.lastTS = pkt.Timestamp
+	b.lastArrival = receivedAt
```
boks1971: Should update this only on RTP time stamp change. As a frame could be broken up into several RTP packets which will all have same RTP timestamp, updating on every packet will skew this a bit towards detecting a burst for a large frame.

boks1971: And also should ignore negative `tsDelta`, i. e. out of order frames.

pkg/pipeline/source/sdk/burst_estimator.go
```diff
@@ -0,0 +1,162 @@
...
+	maxBuffer   int
+}
+
+func newBurstEstimator(clockRate uint32, kind webrtc.RTPCodecType) *burstEstimator {
+	be := &burstEstimator{
+		clockRate:   clockRate,
+		maxSkew:     20 * time.Millisecond,
+		minArrival:  2 * time.Millisecond,
+		scoreTarget: 5,
+		maxBuffer:   1000, // key frames could take up to 600-700 pacejts for hight bitrate streams
```
boks1971: :nit: packets spelling

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -234,16 +236,28 @@ func (w *AppWriter) readNext() {
...
+			}
+		}
+		if !done {
+			return
+		}
+		w.initialized = true
+		packets = ready
+		w.lastReceived.Store(ready[len(ready)-1].ReceivedAt)
+	} else {
+		w.singlePacket[0] = jitter.ExtPacket{ReceivedAt: receivedAt, Packet: pkt}
```
boks1971: Is there a need for this struct field?

boks1971 (approved):
one small comment.

pkg/pipeline/source/sdk/appwriter.go
milos-lk: done to avoid mem allocation for slice - but not really needed - using scalar version of Push API for single packet instead

PR #1033, opened 2025-10-10
Adding linters

milos-lk:
Introducing linting as a first-class build step: build/test/Dockerfile now defines a dedicated golangci stage, copies the prebuilt golangci-lint binary, and wires a lint stage into the pipeline while keeping existing Go module/build caches intact. Build will fail on linter errors and incremental build times should be intact.

Adding .golangci.yaml config with a set of sensible linters - we could enable more but that would require more code changes so I opted to only enable ones which I think are quite useful to have. We could add one by one as we go.

Making few source code changes to satisfy new rules. See comments inline.

### DIFF
--- a/pkg/pipeline/builder/video.go
+++ b/pkg/pipeline/builder/video.go
@@ -233,11 +233,7 @@ func (b *VideoBin) buildWebInput() error {
 		return err
 	}
 
-	if err = b.addDecodedVideoSink(); err != nil {
-		return err
-	}
-
-	return nil
+	return b.addDecodedVideoSink()
 }
 
 func (b *VideoBin) buildSDKInput() error {
@@ -341,16 +337,7 @@ func (b *VideoBin) buildAppSrcBin(ts *config.TrackSource, name string) (*gstream
 			return nil, err
 		}
 
-		if b.conf.VideoDecoding {
-			avDecH264, err := gst.NewElement("avdec_h264")
-			if err != nil {
-				return nil, errors.ErrGstPipelineError(err)
-			}
-
-			if err = appSrcBin.AddElement(avDecH264); err != nil {
-				return nil, err
-			}
-		} else {
+		if !b.conf.VideoDecoding {
 			h264Parse, err := gst.NewElement("h264parse")
 			if err != nil {
 				return nil, errors.ErrGstPipelineError(err)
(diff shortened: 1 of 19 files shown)


pkg/pipeline/builder/video.go
```diff
@@ -341,16 +337,7 @@ func (b *VideoBin) buildAppSrcBin(ts *config.TrackSource, name string) (*gstream
...
-			avDecH264, err := gst.NewElement("avdec_h264")
-			if err != nil {
-				return nil, errors.ErrGstPipelineError(err)
-			}
-
-			if err = appSrcBin.AddElement(avDecH264); err != nil {
-				return nil, err
-			}
-		} else {
+		if !b.conf.VideoDecoding {
```
milos-lk, line 340: early return, avoids superfluous-else

pkg/pipeline/builder/video.go
```diff
@@ -379,17 +375,16 @@ func (b *VideoBin) buildAppSrcBin(ts *config.TrackSource, name string) (*gstream
...
-		if b.conf.VideoDecoding {
-			vp8Dec, err := gst.NewElement("vp8dec")
-			if err != nil {
-				return nil, errors.ErrGstPipelineError(err)
-			}
-			if err = appSrcBin.AddElement(vp8Dec); err != nil {
-				return nil, err
-			}
-		} else {
+		if !b.conf.VideoDecoding {
```
milos-lk, line 378: same as above

pkg/pipeline/builder/video.go
```diff
@@ -407,15 +402,7 @@ func (b *VideoBin) buildAppSrcBin(ts *config.TrackSource, name string) (*gstream
...
-		if b.conf.VideoDecoding {
-			vp9Dec, err := gst.NewElement("vp9dec")
-			if err != nil {
-				return nil, errors.ErrGstPipelineError(err)
-			}
-			if err = appSrcBin.AddElement(vp9Dec); err != nil {
-				return nil, err
-			}
-		} else {
+		if !b.conf.VideoDecoding {
```
milos-lk, line 405: same as above

test/segments.go
```diff
@@ -203,47 +203,64 @@ func (r *Runner) verifySegments(
 	require.Greater(t, segments.Size, int64(0))
 	require.Greater(t, segments.Duration, int64(0))
 
-	r.verifySegmentOutput(t, tc, p, filenameSuffix, segments.PlaylistName, segments.PlaylistLocation, int(segments.SegmentCount), res, m3u8.PlaylistTypeEvent)
+	r.verifySegmentOutput(t, tc, p, filenameSuffix, segmentPlaylist{
```
milos-lk, line 206: too many arguments, added helper struct to encapsulate some of params

PR #1037, opened 2025-10-15
More logs

milos-lk:
To try to figure out issues with startup failures where appsrc report it's not linked before even added.


biglittlebigben (approved):
Lgtm. Errorw may end up too noisy but we can lower the priority later if that happens.

PR #1046, opened 2025-10-26
Detect & handle missing PTS after vp9parse element

milos-lk:
Recently we have observed that a stream corruption can happen in rare cases when a webm muxer receives a buffer stamped by vp9 parser that doesn't have valid PTS. 
Adding probes around vp9parse element so that we can:
- request a key frame when a buffer with no PTS comes out of the vp9 parser. This should hopefully help with healing the stream
- have more debugging info, before we reach out to gstreamer folks as this might be a gst bug (similar [gst issue](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/2509))

### DIFF
--- a/pkg/pipeline/builder/vp9_probe.go
+++ b/pkg/pipeline/builder/vp9_probe.go
@@ -204,9 +204,8 @@ func (p *vp9ParseProbe) requestKeyframeIfDue() {
 		return
 	}
 
-	if p.onSignal != nil {
-		p.onSignal()
-	}
+	p.onSignal()
+
 	p.lastKeyframeRequestNS.Store(now)
 }
 


pkg/pipeline/builder/vp9_probe.go
```diff
@@ -0,0 +1,230 @@
...
+	logger logger.Logger
+
+	lastSrcPTS   atomic.Uint64
+	lastSrcValid atomic.Bool
+	missingPTS   atomic.Bool
+
+	lastSinkPTS   atomic.Uint64
+	lastSinkValid atomic.Bool
+
+	keyframeMu       deadlock.Mutex
```
boks1971: why `deadlock.Mutex`? It is heavier weight right?

pkg/pipeline/builder/vp9_probe.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (p *vp9ParseProbe) onSinkBuffer(_ *gst.Pad, info *gst.PadProbeInfo) gst.PadProbeReturn {
+	buffer := info.GetBuffer()
+	if buffer == nil {
+		return gst.PadProbeOK
+	}
+
+	pts, ok := clockTimeToDuration(buffer.PresentationTimestamp())
+	if !ok {
```
boks1971, line 118: for my own education, why is it okay when there is no timing on sink?

pkg/config/pipeline.go
```diff
@@ -95,6 +95,7 @@ type TrackSource struct {
 	PayloadType     webrtc.PayloadType
 	ClockRate       uint32
 	TempoController *tempo.Controller
+	OnPLIRequired   func()
```
boks1971: suggesting renaming this to `onKeyFrameRequired`.

pkg/pipeline/builder/vp9_probe.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	prev := time.Duration(p.lastSinkPTS.Load())
+	delta := pts - prev
+	if delta < 0 {
+		p.logger.Warnw("vp9parse sink pts moved backwards", nil, "delta", delta)
+		p.logKeyframeHistory("backward_pts")
+	}
+
+	p.lastSinkPTS.Store(uint64(pts))
```
boks1971, line 135: okay, this looks like it is used for getting more information and not force a PLI. Would be good to write a line or two of comments at the top to mention that if that is actually the case.

Also, is it okay to store the moved back PTS here? I guess that is fine. Otherwise, it will keep logging till PTS catches up I guess.

pkg/pipeline/builder/vp9_probe.go
```diff
@@ -0,0 +1,230 @@
...
+
+	return gst.PadProbeOK
+}
+
+func (p *vp9ParseProbe) handleMissingPTS() {
+	if p.onSignal != nil {
```
boks1971: Can we spawn a goroutine here and keep doing this till a key frame is received. We do it every 200ms or so in SFU. Can do the same thing. A single PLI is not guaranteed to produce a key frame.

boks1971: SFU throttles PLIs. And PLIs themselves could technically be lost as they are sent over UDP too. So, technically should not rely on a single PLI.

boks1971 (approved):

pkg/pipeline/builder/vp9_probe.go
milos-lk: maybe you mixed it with.  https://github.com/sasha-s/go-deadlock
This one is: https://pkg.go.dev/github.com/linkdata/deadlock@v0.5.5#section-readme
basically a 0 overhead wrapper the only uses go-deadlock when particular tags are set (ones we enable for test targets). This avoids that fiddling with source files on test runs where we would modify source code to use go-deadlock

boks1971: yeah, I was thinking of that. I did not know about the build tags approach.

Is build tags a robust approach? I am asking specifically because it is 0 overhead only when tags are set. Feels like it is possible to forget setting it. Or somebody else who does not understand it, removing it unintentionally? I wish the default is no overhead and adding a tag makes it a real deadlock detector.

pkg/pipeline/builder/vp9_probe.go
milos-lk: we have option to reject that buffer or opportunistically to let it through - I was reading what vp9 parser usually does in these cases ([example](https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/-/blob/discontinued-for-monorepo/gst/videoparsers/gstvp9parse.c#L424))
Frankly I don't really have strong opinion - was thinking to let it through and downstream element will reject it if they don't like it (like we see they do).

pkg/pipeline/builder/vp9_probe.go
milos-lk: > I am asking specifically because it is 0 overhead only when tags are set

are you sure - this is from their readme?

> The package enables itself when either the deadlock or race build tag is set, and the nodeadlock build tag is not set. The easiest way is to simply use deadlock.(RW)Mutex and run or test your code with the race detector.

pkg/pipeline/builder/vp9_probe.go
milos-lk: > Can we spawn a goroutine here and keep doing this till a key frame is received

sounds good - will do that

pkg/pipeline/builder/vp9_probe.go
boks1971: Got it. Their documentation is confusing for my brain :-). The first part says 

> Changes from that package:

> Uses build tags to eliminate all overhead when not enabled

boks1971: I misread that as build tags need to be set to get 0 overhead. My bad.

pkg/pipeline/builder/vp9_probe.go
milos-lk: updated - started with a separate goroutine but the streaming one (pushing packets) could be used and it's much simpler - added that - please double check 🙏

pkg/pipeline/builder/vp9_probe.go
```diff
@@ -0,0 +1,262 @@
...
+		return
+	}
+
+	now := time.Now().UnixNano()
+	last := p.lastKeyframeRequestNS.Load()
+	if last != 0 && time.Duration(now-last) < keyframeRequestInterval {
+		return
+	}
+
+	if p.onSignal != nil {
```
boks1971: Guess there is no need to check for nil again here

pkg/pipeline/builder/vp9_probe.go
```diff
@@ -0,0 +1,262 @@
...
+			p.totalIntervalSum += delta
+			p.totalIntervals++
+		}
+	}
+
+	p.keyframePTS = append(p.keyframePTS, pts)
+	if len(p.keyframePTS) > keyframeHistorySize {
+		p.keyframePTS = p.keyframePTS[1:]
+		// sliding window only keeps the most recent timestamps for debugging logs
+	}
```
boks1971, line 189: Should thus be clearing key range pending. Did not see it getting cleared. Possible I missed it.

milos-lk: We set it to false on line 166

pkg/pipeline/builder/vp9_probe.go
milos-lk: ugh - forgot to remove this 👍

pkg/pipeline/builder/vp9_probe.go
boks1971: Missed it. Thank you.

Issue #956, opened 2025-07-31
[BUG] {publisher_identity} templating not working for segment output


aureus1:
**Describe the bug**
`{publisher_identity}` is not replaced when using `segment_outputs.filename_prefix`

<img width="412" height="462" alt="Image" src="https://github.com/user-attachments/assets/fcacf4f6-f937-41af-866e-9fc4cdd7ac80" />

**Egress Version**
`Image:          livekit/egress:v1.8.4`

**Egress Request**
Using participant auto-egress with the following json:
```
{
      "file_outputs": [{
        "file_type": "MP4",
        "filepath": "livekit-poc-auto/{room_name}/{time}-{publisher_identity}",
        "s3": {
          "region": "us-east-1",
          "bucket": "livekit-egress-buckett"
        }
      }],
      "segment_outputs": [
        {
          "filename_prefix": "livekit-poc-auto/{room_name}/hls/{time}-{publisher_identity}/",
          "playlist_name": "playlist.m3u8",
          "live_playlist_name": "playlist-live.m3u8",
          "segment_duration": 10,
          "s3": {
            "region": "us-east-1",
            "bucket": "livekit-egress-buckett"
          }
        }
      ]
}
```

**Additional context**
[Documentation ](https://docs.livekit.io/home/egress/outputs/#filename-templating)indicates that the `{publisher_identity}` token should be replaced for filename_prefix

**Logs**

PR #1051, opened 2025-11-03
Fix filename templating in case of multi output

milos-lk:
Due to a premature return if muliple outputs are specified, for all outputs after the one for file, filename templates aren't replaced.

The change fixes it by alowing processing all outputs before returning. Fix for:
https://github.com/livekit/egress/issues/956

### DIFF
--- a/pkg/config/pipeline.go
+++ b/pkg/config/pipeline.go
@@ -592,13 +592,14 @@ func (p *PipelineConfig) getRoomCompositeRequestType(req *livekit.RoomCompositeE
 
 // used for sdk input source
 func (p *PipelineConfig) UpdateInfoFromSDK(identifier string, replacements map[string]string, w, h uint32) error {
+	var err error
 	for egressType, c := range p.Outputs {
 		if len(c) == 0 {
 			continue
 		}
 		switch egressType {
 		case types.EgressTypeFile:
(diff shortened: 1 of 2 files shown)


pkg/config/pipeline.go
```diff
@@ -592,13 +592,14 @@ func (p *PipelineConfig) getRoomCompositeRequestType(req *livekit.RoomCompositeE
...
 func (p *PipelineConfig) UpdateInfoFromSDK(identifier string, replacements map[string]string, w, h uint32) error {
+	var err error
 	for egressType, c := range p.Outputs {
 		if len(c) == 0 {
 			continue
 		}
 		switch egressType {
 		case types.EgressTypeFile:
-			return c[0].(*FileConfig).updateFilepath(p, identifier, replacements)
+			err = c[0].(*FileConfig).updateFilepath(p, identifier, replacements)
```
boks1971, line 602: can do uber/multierr and append all errors. Good to combine errors.

boks1971 (approved):

pkg/config/pipeline.go
milos-lk: could be handy - though maybe not for this case - as we support only one file output per egress. noted

biglittlebigben (approved):

PR #1059, opened 2025-11-10
Do not consider disabled egresses as idle

biglittlebigben:


biglittlebigben:
> will this cause a legitimately disabled region to scale up or down immediately?

It would, but this will not actually affect cloud. Cloud PR incoming.

PR #1080, opened 2025-11-26
Adding ability for configuring max file size in bytes as a storage quota

milos-lk:
Since storage is limited resource, adding ability to add max file size quota for file recordings. The flow and the final state are similar to what's already present for FileOutputMaxDuration.
In case the file size exceeds the configured max value `EgressStatus_EGRESS_LIMIT_REACHED` is returned.

### DIFF
--- a/pkg/pipeline/controller.go
+++ b/pkg/pipeline/controller.go
@@ -565,10 +565,6 @@ func (c *Controller) startSessionLimitTimer(ctx context.Context) {
 }
 
 func (c *Controller) startOutputSizeMonitor() {
-	if c.storageMonitorCancel != nil {
-		return
-	}
-
 	ctx, cancel := context.WithCancel(context.Background())
 	c.storageMonitorCancel = cancel
 


pkg/pipeline/controller.go
```diff
@@ -548,6 +564,153 @@ func (c *Controller) startSessionLimitTimer(ctx context.Context) {
 	}
 }
 
+func (c *Controller) startOutputSizeMonitor() {
+	if c.storageMonitorCancel != nil {
```
boks1971: can this race and create two cancels? Does this need to be protected?

pkg/pipeline/controller.go
```diff
@@ -548,6 +564,153 @@ func (c *Controller) startSessionLimitTimer(ctx context.Context) {
...
+			return
+		}
+
+		thresholdTriggered := false
+		for nextThreshold < len(thresholds) && size >= thresholds[nextThreshold] {
+			logger.Debugw(
+				"output size threshold exceeded",
+				"dir", c.TmpDir,
+				"bytesWritten", size,
+				"thresholdBytes", thresholds[nextThreshold],
```
boks1971, line 638: it is useful to list possible top 10 files/sizes so that it is easier to see in logs what is causing the bloat? i. e. log file sizes at thresholds also and not just when the limit is reached.

milos-lk: the list of files is already sorted - the biggest ones should appear first - we can take top 10 anyways - it could reduce verbosity in logs 👍 

> i. e. log file sizes at thresholds also and not just when the limit is reached.

I think I am logging file sizes at these thresholds: https://github.com/livekit/egress/blob/9fffa2b0f8c9059741f41e4e8046196e0f2e6b2b/pkg/pipeline/controller.go#L648 
or I didn't understand the question?

pkg/pipeline/controller.go
milos-lk: we call Run on controller only once - so there is no potential for the race - but this check isn't that useful. It's just me being paranoid a bit :)

pkg/pipeline/controller.go
milos-lk: updated to log top 10 only

boks1971: Sorry, missed the logging below. Yeah, that is what I meant, but missed the logging message below.

pkg/pipeline/controller.go
milos-lk: I am going to delete the check - as it might be confusing

milos-lk: updated

pkg/config/base.go
```diff
@@ -68,6 +68,7 @@ type BaseConfig struct {
 
 type SessionLimits struct {
 	FileOutputMaxDuration    time.Duration `yaml:"file_output_max_duration"`
+	FileOutputMaxSize        int64         `yaml:"file_output_max_size"` // max on-disk size in bytes before stopping; 0 to disable
```
biglittlebigben, line 71: The naming of this conf variable is a bit confusing now that we check the full size of the handler storage, including temp files.

milos-lk: What do you think about `MaxStorageSize`?

biglittlebigben: Sure

PR #1090, opened 2025-12-18
Drain appwriter

milos-lk:
When Drain() was called on an AppWriter, not all samples buffered in the jitter buffer and internal channel were being pushed to the pipeline before the stream ended. This resulted in possibly lost audio/video samples at the end of tracks.

Replaced the single endStream fuse with three separate fuses to track different stages of the drain process:
  - endStreamSignaled: Drain/end has been requested
  - endStreamSourceProcessed: Jitter buffer has been flushed and all packets extracted
  - endStreamProcessed: All samples have been pushed to the pipeline

Added new Flush API on jitter buffer to drain the remaining samples as well at the end.

### DIFF
--- a/pkg/pipeline/source/sdk/appwriter.go
+++ b/pkg/pipeline/source/sdk/appwriter.go
@@ -46,6 +46,7 @@ const (
 	discontinuityTolerance = 500 * time.Millisecond
 	pipelineCheckInterval  = 5 * time.Second
 	cSamplesQueueDepth     = 100
+	drainingTimeout        = time.Second * 3
 )
 
 type sampleItem struct {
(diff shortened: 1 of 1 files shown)


frostbyte73 (approved):

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -220,9 +222,13 @@ func (w *AppWriter) start() {
 	}()
 
 	go w.pushSamples()
-	for !w.endStream.IsBroken() {
+	for !w.endStreamSignaled.IsBroken() {
 		w.readNext()
 	}
+	w.drainJitterBuffer()
+
+	<-w.endStreamProcessed.Watch()
```
biglittlebigben: Should we add a timeout here? Even if we can't think of a case where we may have deadlock currently, unexpected things can happen and a future change may change the flow.

milos-lk: doesn't hurt to be extra defensive - will add

milos-lk: added the timeout - if something goes wrong and we hit it - the goroutine won't wait forever and will buy us not having pipeline frozen error but if it happens data races are possible - this is something we should make we know if it happens (error log added).

PR #1093, opened 2025-12-19
Handle EOS when removing source bin

milos-lk:
Making sure last bits of source bin produced data arrive at the sink bin (including the EOS event). This prevents minor data loss when the source bin gets removed. This is not hard deadline - removal process hasn't been made sync, installed probes will just pass data bit longer (until EOS passes). Catch all timeout is also added so the removal happens anyways.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -17,6 +17,7 @@ linters:
     - fatcontext # detects nested contexts in loops
     - goconst # checks for hardcoded strings that could be replaced with a constant
     - misspell # checks for misspelled words
+    - staticcheck
 
 linters-settings:
   revive:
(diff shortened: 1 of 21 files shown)


pkg/gstreamer/pads.go
```diff
@@ -117,7 +127,7 @@ func matchPadsLocked(src, sink *Bin) (*gst.Pad, *gst.Pad, error) {
 		return srcPad, sinkTemplates[0].toPad(), nil
 	case sinkPad != nil && len(srcTemplates) == 1:
 		return srcTemplates[0].toPad(), sinkPad, nil
-	case len(srcTemplates) >= 1 && len(srcTemplates) >= 1:
+	case len(srcTemplates) > 0 && len(sinkTemplates) > 0:
```
milos-lk, line 130: this one and one below are just side fixes I picked up while changing things here.

PR #1094, opened 2025-12-19
Enable staticcheck

milos-lk:
Run into some logic errors which could have been caught by staticcheck has it been enabled. So maybe a good time to strengthen the list of linters we have enabled by adding it and fixing outlined warnings.


pkg/handler/handler.go
```diff
@@ -88,7 +88,7 @@ func NewHandler(conf *config.PipelineConfig, bus psrpc.MessageBus) (*Handler, er
 }
 
 func (h *Handler) Run() {
-	ctx, span := tracer.Start(context.Background(), "Handler.Run")
+	ctx, span := otel.Tracer("egress.handler").Start(context.Background(), "Handler.Run")
```
boks1971: is that supposed to be the package name given to `otel.Tracer`?

pkg/ipc/conn.go
```diff
@@ -53,7 +53,7 @@ func StartServiceListener(ipcServer *grpc.Server, serviceTmpDir string) error {
 
 func NewHandlerClient(handlerTmpDir string) (*EgressHandlerClientWrapper, error) {
 	socketAddr := path.Join(handlerTmpDir, handlerAddress)
-	conn, err := grpc.Dial(socketAddr,
+	conn, err := grpc.NewClient(socketAddr,
```
boks1971, line 55: How does `Dial` and `NewClient` differ?

milos-lk: they have one subtle difference which caused all tests to fail here:
```
// One subtle difference between NewClient and Dial and DialContext is that the
// former uses "dns" as the default name resolver, while the latter use
// "passthrough" for backward compatibility.  This distinction should not matter
// to most users, but could matter to legacy users that specify a custom dialer
// and expect it to receive the target string directly.
```
We don't need custom dialer, removed it should be completely equivalent

pkg/handler/handler.go
milos-lk: based on the doc:
```
// The passed name needs to uniquely identify instrumentation code.
// Therefore, it is recommended that name is the Go package name of the library providing instrumentation (note: not the code being instrumented)
```
hm this needs to be something like `otel` reading it again... 🤔 
not sure would need to check

PR #1096, opened 2025-12-22
Log file upload stats (size, duration)

milos-lk:
In some edge cases file upload operation could take a while - adding more logging to be able to isolate these and check upload bitrate.

### DIFF
--- a/pkg/pipeline/sink/file.go
+++ b/pkg/pipeline/sink/file.go
@@ -90,7 +90,7 @@ func (s *FileSink) Close() error {
 
 	s.FileInfo.Location = location
 	s.FileInfo.Size = size
-	logger.Infow("file upload completed",
+	logger.Debugw("file upload completed",
 		"bytes", size,
 		"duration", time.Since(start))
 


pkg/pipeline/sink/file.go
```diff
@@ -79,13 +81,20 @@ func (s *FileSink) UploadManifest(filepath string) (string, bool, error) {
...
 	location, size, err := s.Upload(s.LocalFilepath, s.StorageFilepath, s.OutputType, false)
 	if err != nil {
+		logger.Debugw("file upload failed", err)
 		return err
 	}
 
 	s.FileInfo.Location = location
 	s.FileInfo.Size = size
+	logger.Infow("file upload completed",
+		"storagePath", s.StorageFilepath,
```
boks1971: is this okay to log? will this have PII?

milos-lk: actually - we already have some file into when we log validated requests - these are not needed - I will omit them here just in case.

PR #1105, opened 2026-01-13
Support for mulaw and alaw input codecs

milos-lk:
Egress support for recording input tracks coming encoded in G711 codecs.

Fairly straight forward change - adding new depay/decoder elements depending on the codec type. Some changes were needed to enable reliable testing - namely creating new rooms for these codecs and setting G711 codecs for test publishers explicitly so there are some changes in testing code around that.


pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -675,3 +679,21 @@ func (w *AppWriter) ensureRemovedBeforeDrain() {
 		w.callbacks.OnTrackRemoved(w.track.ID())
 	}
 }
+
+type G711Packet struct{}
```
milos-lk, line 683: pion doesn't seem to have it - so added here - but probably needs new home

biglittlebigben (approved):

biglittlebigben (approved):

PR #1111, opened 2026-01-27
Make sure all data is read from LK server

milos-lk:
OnTrackUnsubscribed doesn't come in band with media data, some media data might still be buffered on SFU side when egress proceeds with reacting on this signal by stopping its read loop and draining all internally buffered data. This might result in missing to read last buffered samples on the SFU side. 
The goal of this change is to try to read all data from SFU - even after `OnTrackUnsubscribed` is received by not stopping the read loop and allowing up to grace period to read remaining media. Instead of stopping, keep record of the received unsubscribed signal and keep reading until either EOF or an unexpected error is received or grace period expires - at which point data will be drained as usual.

### DIFF
--- a/test/builder.go
+++ b/test/builder.go
@@ -27,7 +27,7 @@ import (
 	"github.com/livekit/protocol/utils"
 )
 
-const webUrl = "https://storage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"
+const webUrl = "https://download.blender.org/peach/bigbuckbunny_movies/BigBuckBunny_320x180.mp4"
 
 type testCase struct {
 	name        string


pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -321,17 +323,38 @@ func (w *AppWriter) handleReadError(err error) {
 	var netErr net.Error
 	switch {
 	case w.draining.IsBroken():
+		if !w.endStreamSignaled.IsBroken() {
+			// Delayed drain in progress (Drain(false) was called, timer pending)
+			if (errors.As(err, &netErr) && netErr.Timeout()) || err.Error() == errBufferTooSmall {
```
boks1971, line 328: Should this continue reading on buffer too small?

milos-lk: that's a good question - I took the default handling of that error below which just logs error and continues. tbh - I am not sure under which circumstances that error could occur - to me it looks like it's a guard for malformed/huge packets? (@frostbyte73 or @biglittlebigben could you maybe double check). I think ignoring it is the right thing to do wrt resilience.

biglittlebigben: Yes, this is an error returned by pion in case something went wrong in the parsing. Treating this as a non fatal error is probably right, but it's hard to be certain considering this is not an expected situation...

PR #1116, opened 2026-02-02
Instrument all leaky queues

milos-lk:
Recently we have added instrumentation for video leaky queues. However this concept is also useful for audio. These changes are installing instrumentation for all leaky queues.

### DIFF
--- a/pkg/pipeline/controller.go
+++ b/pkg/pipeline/controller.go
@@ -80,10 +80,10 @@ type Controller struct {
 }
 
 type controllerStats struct {
-	droppedAudioBuffers atomic.Uint64 // dropped by mixer
-	droppedVideoBuffers atomic.Uint64
+	mixerDroppedAudioBuffers atomic.Uint64
+	droppedVideoBuffers      atomic.Uint64
 
-	droppedAudioDuration atomic.Duration // dropped by mixer
+	mixerDroppedAudioDuration atomic.Duration
 
 	queuesDroppedAudioBuffers atomic.Uint64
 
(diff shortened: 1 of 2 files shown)


pkg/pipeline/controller.go
```diff
@@ -80,9 +80,14 @@ type Controller struct {
 }
 
 type controllerStats struct {
-	droppedAudioBuffers      atomic.Uint64
-	droppedAudioDuration     atomic.Duration
-	droppedVideoBuffers      atomic.Uint64
+	droppedAudioBuffers atomic.Uint64 // dropped by mixer
+	droppedVideoBuffers atomic.Uint64
+
+	droppedAudioDuration atomic.Duration // dropped by mixer
```
boks1971: maybe prefix (or suffix) these with `mixer` just like the `queues` prefix 👇 ?

milos-lk: done

pkg/pipeline/controller.go
```diff
@@ -225,8 +225,8 @@ func (c *Controller) Run(ctx context.Context) *livekit.EgressInfo {
 		if c.SourceType == types.SourceTypeSDK {
 			logger.Debugw(
 				"audio qos stats",
-				"audioBuffersDropped", c.stats.droppedAudioBuffers.Load(),
-				"totalAudioDurationDropped", c.stats.droppedAudioDuration.Load(),
+				"audioBuffersDropped", c.stats.mixerDroppedAudioBuffers.Load(),
```
boks1971, line 228: maybe logging keys here as well to indicate `mixer` as source of drop?

milos-lk: I have some dashboards built based on this which would need adjustments (temp duplication) to handle both values - maybe I will just duplicate the field for now - add the new one and remove the old in the subsequent release.

boks1971: I think it is okay to leave as is. Probably not worth the double change and all the side-effects.

PR #1118, opened 2026-02-02
 Add cgroup-aware memory monitoring for admission control and OOM kill 

milos-lk:
## Summary 
                                                                            
  - Add configurable memory source (`proc_rss` or `cgroup`) for admission control and kill decisions, using   
  `hwstats.MemoryStats` from livekit protocol for cgroup reads                                                
  - When `memory_source: cgroup` is configured, use cgroup working set memory instead of per-process RSS sums 
  for more accurate container-aware memory accounting                                                         
  - Graceful fallback to `proc_rss` when cgroup stats are unavailable, with throttled error logging           
  - Add configurable kill grace period (`memory_kill_grace_sec`) to avoid premature kills on transient spikes 
  - Add Prometheus metrics for cgroup memory, proc RSS, read success, and "would reject" evaluation gauges  to be  able to evaluate how would cgroup based memory stats influence admission control without even switching to it
  - Default behavior is unchanged (`proc_rss`) — fully backward compatible                                    
                                                                                                              
 ## Motivation                       
 
Per-process RSS sums can undercount actual container memory usage (shared pages, kernel buffers, page cache) and miss memory pressure that leads to OOM kills at the container/cgroup level. Using cgroup memory stats provides a more accurate view of memory consumption inside Kubernetes pods and Docker containers.

### DIFF
--- a/pkg/stats/monitor.go
+++ b/pkg/stats/monitor.go
@@ -69,15 +69,15 @@ type Monitor struct {
 	pendingPulseClients atomic.Int32
 	pendingMemoryUsage  atomic.Float64
 
-	mu                 deadlock.Mutex
-	highCPUDuration    int
-	highMemoryDuration int
-	pending            map[string]*processStats
-	procStats          map[int]*processStats
-	memoryUsage        float64
-	cgroupUsageBytes   uint64
-	cgroupOK           bool
-	cgroupErrorLogged  atomic.Bool
+	mu                deadlock.Mutex
+	highCPUDuration   int
+	highMemoryStart   time.Time
+	pending           map[string]*processStats
+	procStats         map[int]*processStats
+	memoryUsage       float64
+	cgroupUsageBytes  uint64
+	cgroupOK          bool
+	cgroupErrorLogged atomic.Bool
 }
 
 type processStats struct {
@@ -659,10 +659,10 @@ func (m *Monitor) checkMemoryKill(maxMemoryEgress string) {
 
 	if killTriggerBytes > maxMemoryBytes {
 		// Apply grace period if configured.
-		// Note: highMemoryDuration counts update cycles (typically 1 second each),
-		// so MemoryKillGraceSec is approximate.
-		m.highMemoryDuration++
-		if m.highMemoryDuration > m.cpuCostConfig.MemoryKillGraceSec {
+		if m.highMemoryStart.IsZero() {
+			m.highMemoryStart = time.Now()
+		}
+		if time.Since(m.highMemoryStart) >= time.Duration(m.cpuCostConfig.MemoryKillGraceSec)*time.Second {
 			killTriggerGB := float64(killTriggerBytes) / gb
 			logger.Warnw("high memory usage", nil,
 				"source", m.cpuCostConfig.MemorySource,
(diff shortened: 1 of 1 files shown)


pkg/stats/monitor.go
```diff
@@ -528,11 +555,98 @@ func (m *Monitor) updateEgressStats(stats *hwstats.ProcStats) {
...
+	}
+
+	m.cgroupOK = true
+	m.cgroupErrorLogged.Store(false)
+	m.cgroupUsageBytes = usageBytes
+
+	m.promCgroupReadSuccess.Set(1)
+	m.promCgroupMemory.Set(float64(usageBytes))
+	logger.Infow("cgroup memory stats", "usageBytes", usageBytes, "limitBytes", limitBytes)
+	logger.Infow("rss memory stats", "memoryUsage", m.memoryUsage)
```
milos-lk, line 621: would like to have these for a few test runs on staging - will remove them in the subsequent PR.

pkg/stats/monitor.go
```diff
@@ -528,11 +583,98 @@ func (m *Monitor) updateEgressStats(stats *hwstats.ProcStats) {
...
+			killTriggerBytes = m.cgroupUsageBytes
+		}
+	default: // proc_rss
+	}
+
+	if killTriggerBytes > maxMemoryBytes {
+		// Apply grace period if configured.
+		// Note: highMemoryDuration counts update cycles (typically 1 second each),
+		// so MemoryKillGraceSec is approximate.
+		m.highMemoryDuration++
```
boks1971: Is this checking every second? the `++` and a check for grace seconds below threw me of. If it is purely time based, can it just use `time` rather than a counter?

milos-lk: yes - wanted to modify this but forgot - will update it to be time based.

milos-lk:
> I am not sure where the RSS numbers are coming from. But, wondering if that can be folded into `protocol` `hwstats` module as an option so that other services can also potentially use it?

It's already there:
https://github.com/livekit/protocol/blob/main/utils/hwstats/cpu.go#L70

pkg/stats/monitor.go
milos-lk: done

pkg/stats/monitor.go
```diff
@@ -528,11 +583,98 @@ func (m *Monitor) updateEgressStats(stats *hwstats.ProcStats) {
 	}
 
 	m.memoryUsage = float64(totalMemory) / gb
-	if m.cpuCostConfig.MaxMemory > 0 && totalMemory > int(m.cpuCostConfig.MaxMemory*gb) {
-		logger.Warnw("high memory usage", nil,
-			"memory", m.memoryUsage,
-			"requests", m.requests.Load(),
-		)
-		m.svc.KillProcess(maxMemoryEgress, errors.ErrOOM(float64(maxMemory)/gb))
+	m.promProcRSS.Set(float64(totalMemory))
```
biglittlebigben, line 586: Why storing this twice?

milos-lk: promProcRSS is a metric we can plot to get the RSS usage tracked (as seen by process). Or I missed something?

biglittlebigben: 👍

PR #1125, opened 2026-02-06
Fix track resubscription race condition with per-track state machine

milos-lk:
## Problem
When a publisher disconnects/reconnects or is migrated, the track gets unsubscribed and resubscribed with the same trackID. A race condition occurs because writer cleanup could take up to several seconds (draining SFU buffers, grace period for late packets), and the new subscription can arrive while the old writer is still active.

IMPORTANT: this builds on top of https://github.com/livekit/egress/pull/1111 and is its prerequisite 

### The race:
  1. Subscribe(T1) → W1 created, stored in map
  2. Unsubscribe(T1) → W1 starts cleanup (async, takes seconds)
  3. Subscribe(T1) → W2 created, overwrites W1 in map
  4. W1 cleanup done → deletes W2 from map, corrupts synchronizer

  Symptoms: map corruption, synchronizer corruption, GStreamer element naming collisions, resource leaks

## Solution

  A per-track state machine that queues pending subscriptions during cleanup instead of overwriting.

  States: IDLE → ACTIVE → CLEANING → IDLE (or back to ACTIVE if pending exists)

  Key Changes

  - trackState struct — tracks writer, pending subscription, and cleaning flag per track
  - storeWriter() — stores writer atomically, calls active.Inc() only on success
  - tryQueuePendingSubscription() — queues subscription when state changes during writer creation
  - handleOrphanedWriter() — cleans up writers that couldn't be stored due to race
  - subscriptionQueued flag — prevents false success reports to awaitExpected
  - Two-phase onTrackFinished() — keeps cleaning=true during cleanup, executes pending after

  ---
  Edge Cases Handled

  - Subscribe while draining — queued, executed after cleanup
  - Unsubscribe during cleanup — clears pending (cancels queued subscription)
  - Rapid flapping — last event wins, writer identity check prevents stale cleanup
  - State changes during writer creation — queued, result suppressed until executed
  - Shutdown during creation — dropped gracefully, no false success

## Alternative solution
Allowing 2 tracks for the same track ID to co-exist - much simpler - but the bad part is that it relies on the pion track pointers comparison for uniquely identifying tracks: https://github.com/livekit/egress/pull/1117

### DIFF
--- a/pkg/pipeline/source/sdk.go
+++ b/pkg/pipeline/source/sdk.go
@@ -43,6 +43,18 @@ const (
 	subscriptionTimeout = time.Second * 30
 )
 
+type trackState struct {
+	writer   *sdk.AppWriter  // current active writer (nil if none)
+	pending  *pendingSubData // queued subscription while cleaning
+	cleaning bool            // cleanup in progress
+}
+
+type pendingSubData struct {
+	track *webrtc.TrackRemote
+	pub   *lksdk.RemoteTrackPublication
+	rp    *lksdk.RemoteParticipant
+}
+
 type SDKSource struct {
 	*config.PipelineConfig
 	callbacks *gstreamer.Callbacks
@@ -55,7 +67,9 @@ type SDKSource struct {
 	filenameReplacements map[string]string
 	subs                 chan *subscriptionResult
 
-	writers map[string]*sdk.AppWriter
+	trackStates map[string]*trackState
+	closing     bool
+
 	subLock deadlock.RWMutex
 	active  atomic.Int32
 	closed  core.Fuse
@@ -80,7 +94,8 @@ func NewSDKSource(ctx context.Context, p *config.PipelineConfig, callbacks *gstr
 		callbacks:            callbacks,
 		filenameReplacements: make(map[string]string),
 		subs:                 make(chan *subscriptionResult, 100),
-		writers:              make(map[string]*sdk.AppWriter),
+		trackStates:          make(map[string]*trackState),
+		closing:              false,
 	}
 	logger.Debugw("latency config", "latency", p.Latency)
 
@@ -138,9 +153,13 @@ func (s *SDKSource) EndRecording() <-chan struct{} {
 }
 
 func (s *SDKSource) Playing(trackID string) {
-	s.mu.Lock()
-	writer := s.writers[trackID]
-	s.mu.Unlock()
+	s.mu.RLock()
+	state := s.trackStates[trackID]
+	var writer *sdk.AppWriter
+	if state != nil {
+		writer = state.writer
+	}
+	s.mu.RUnlock()
 
 	if writer != nil {
 		writer.Playing()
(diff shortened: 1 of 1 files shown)


pkg/pipeline/source/sdk.go
```diff
@@ -43,6 +43,18 @@ const (
 	subscriptionTimeout = time.Second * 30
 )
 
+type trackState struct {
+	writer   *sdk.AppWriter  // current active writer (nil if none)
+	pending  *pendingSubData // queued subscription while cleaning
```
boks1971, line 48: Maybe overkill, but I would make this a slice/list and just handle it as chain of pending operations to make it generic.

boks1971: Suggest adding a `generation` field too and log them when de-queuing a subscription so that it is easy to see how many times this is happening.

pkg/pipeline/source/sdk.go
```diff
@@ -473,6 +527,29 @@ func (s *SDKSource) subscribe(track lksdk.TrackPublication) error {
...
+	s.mu.Lock()
+	if s.closing {
+		s.mu.Unlock()
+		logger.Debugw("ignoring subscription, source closing", "trackID", trackID)
+		return
+	}
+
+	state := s.trackStates[trackID]
+	if state != nil && state.cleaning {
+		state.pending = &pendingSubData{track: track, pub: pub, rp: rp}
```
boks1971, line 541: based on slice comment above, this would be an append.

pkg/pipeline/source/sdk.go
```diff
@@ -693,13 +815,29 @@ func (s *SDKSource) onTrackUnsubscribed(_ *webrtc.TrackRemote, pub *lksdk.Remote
...
+	}
+	state := s.trackStates[trackID]
+	if state == nil || state.writer == nil {
+		logger.Warnw("ignoring track unsubscribed, writer not found", nil, "trackID", trackID)
+		s.mu.Unlock()
+		return
+	}
 
-	if writer == nil {
+	if state.cleaning {
```
boks1971, line 830: guess, the unsubscribe will become an op to be queued if changing to slice of ops.

pkg/pipeline/source/sdk.go
```diff
@@ -709,43 +847,71 @@ func (s *SDKSource) onTrackUnsubscribed(_ *webrtc.TrackRemote, pub *lksdk.Remote
...
+		s.mu.Unlock()
+		return
 	}
+	// After this unlock, new subscriptions create writers directly (won't queue)
+	pending := state.pending
+	state.cleaning = false
+	state.pending = nil
+	s.mu.Unlock()
+
+	if pending != nil && !s.closing {
```
boks1971, line 910: s.closing accessed outside the lock, is that okay?

pkg/pipeline/source/sdk.go
milos-lk: nice idea for making this pending a slice and handling all of operations there - though I need to think a bit about that because could blow complexity out of proportion. Did you have in mind a generic serial queue of all operations involving writer state (not only subsribed/unsubsribed)?

pkg/pipeline/source/sdk.go
milos-lk: good catch - this should be read in the critical section

pkg/pipeline/source/sdk.go
boks1971: Yeah, I was think about a queue of ops. Possible ops 
- subscribe
- unsubscribe
- finished

And a state machine if it can be in only one state. Currently, it looks like two things (`active` counter and `cleaning` flag - so it may not fit well). But, I was actually expecting these states when I read the PR commit notes
```
States: IDLE → ACTIVE → CLEANING → IDLE (or back to ACTIVE if pending exists)
```
From that transition, I thought `active` and `cleaning` need not be separate variables. I always find it easier to reason with a single `state` variable going through those transitions.

Basically, a nested switch makes it easier for me to reason and not miss conditions. So, something like
```
switch (currentState) {
    case "IDLE":
        switch (op) {
            case "SUBSCRIBE":
               doRelevantWork()
               
            case "UNSUBSCRIBE":
               -> this is an invalid transition
         }

    case "ACTIVE":
         switch (op) {
         }

   ....
}
```
That gives me a clear picture of all possible transitions and also all illegal transitions very easy to see.

milos-lk: I like that - I think I have an idea on how to do it - will create a separate PR - I hope it won't turn the sdk.go upside down :)

PR #1127, opened 2026-02-11
Explicit per track FSM for track/writer operations

milos-lk:
Refactor track subscription handling to use an explicit per-track state machine, supersedes https://github.com/livekit/egress/pull/1125
Replacing implicit state (scattered fields like writers map, cleaning flags etc) with explicit states (IDLE/ACTIVE/CLEANING). Each track gets its own worker goroutine to serialize operations (they are all sequentially processed) and avoid cross-track blocking - this helps ensuring callbacks invoked from the sdk from different goroutines are serialized for the same trackID - so track local state doesn't need locks and all state transitions could be performed fully synchronously (no inter track blocking on e.g draining etc). Generation tracking prevents stale Playing() calls from affecting new writers after resubscribe.

```mermaid
graph TB                                                                                 
      subgraph "LiveKit SDK Callbacks"   
          OTS[onTrackSubscribed]                                                           
          OTU[onTrackUnsubscribed]                                              
          OTF[onTrackFinished]                                                             
      end                                                   

      subgraph "SDKSource"                                                      
          SO[submitOp]
          GOC[getOrCreateWorker]
          WM["workers map[trackID]*trackWorker"]
      end

      subgraph "Per-Track Workers"
          subgraph "Worker T1"
              W1[runWorker goroutine]
              WS1["workerState
              • state: TrackState
              • writer: *AppWriter
              • generation: uint64"]
              OC1["opChan (buffered)"]
          end

          subgraph "Worker T2"
              W2[runWorker goroutine]
              WS2["workerState
              • state: TrackState
              • writer: *AppWriter
              • generation: uint64"]
              OC2["opChan (buffered)"]
          end
      end

      subgraph "Shared Resources"
          SYNC[Synchronizer]
          CB[Callbacks]
          ACT["active (atomic.Int32)"]
      end

      OTS -->|OpSubscribe| SO
      OTU -->|OpUnsubscribe| SO
      OTF -->|OpFinished| SO

      SO --> GOC
      GOC --> WM
      WM --> OC1
      WM --> OC2

      OC1 --> W1
      OC2 --> W2

      W1 --> WS1
      W2 --> WS2

      W1 --> SYNC
      W1 --> CB
      W1 --> ACT
      W2 --> SYNC
      W2 --> CB
      W2 --> ACT
```

FSM

```mermaid
stateDiagram-v2
      [*] --> IDLE: Worker created

      IDLE --> ACTIVE: OpSubscribe (create writer)
      IDLE --> [*]: OpClose

      ACTIVE --> CLEANING: OpUnsubscribe
      ACTIVE --> CLEANING: OpFinished
      ACTIVE --> [*]: OpClose (drain + exit)

      CLEANING --> IDLE: Cleanup complete

      note right of IDLE: No writer, ops ignored
      note right of ACTIVE: Writer running, generation tracked
      note right of CLEANING: Blocking drain, ops queued
```

### DIFF
--- a/pkg/pipeline/source/sdk.go
+++ b/pkg/pipeline/source/sdk.go
@@ -53,10 +53,9 @@ type SDKSource struct {
 	workersMu deadlock.RWMutex
 	workers   map[string]*trackWorker
 
-	// subLock ensures in-flight pre-init subscriptions complete before initialized.Break().
-	// This keeps pre-init errors in the startup failure path rather than OnError path.
-	// Workers hold RLock during subscription processing.
-	// Await functions hold Lock before calling initialized.Break().
+	// subLock prevents a race where a subscription starts during init completion.
+	// Without it, the subscription could see "not yet initialized", then init completes,
+	// leaving the track orphaned (missed by both pipeline build and dynamic add).
 	subLock deadlock.RWMutex
 
 	closing atomic.Bool
@@ -303,11 +302,10 @@ func (s *SDKSource) stopAwaitingTracks() {
 	s.initResultChan.Store(nil) // just nil out, don't close
 }
 
-// waitForInFlightSubscriptions blocks until all workers release their RLock.
-// This ensures in-flight subscription processing completes before initialized.Break(),
-func (s *SDKSource) waitForInFlightSubscriptions() {
+func (s *SDKSource) completeInit() {
 	s.subLock.Lock()
-	s.subLock.Unlock() //nolint:staticcheck // SA2001: intentionally empty - Lock is a barrier
+	defer s.subLock.Unlock()
+	s.initialized.Break()
 }
 
 // getInitResultChan returns the current init result channel (nil after init complete)
@@ -345,8 +343,7 @@ func (s *SDKSource) awaitRoomTracks() error {
 		return err
 	}
 
-	s.waitForInFlightSubscriptions()
-	s.initialized.Break()
+	s.completeInit()
 	return nil
 }
 
@@ -377,8 +374,7 @@ func (s *SDKSource) awaitParticipantTracks(identity string) (uint32, uint32, err
 		}
 	}
 
-	s.waitForInFlightSubscriptions()
-	s.initialized.Break()
+	s.completeInit()
 	return w, h, nil
 }
 
@@ -423,8 +419,7 @@ func (s *SDKSource) getParticipant(identity string) (*lksdk.RemoteParticipant, e
 func (s *SDKSource) awaitTracks(expecting map[string]struct{}) (uint32, uint32, error) {
 	trackCount := len(expecting)
 	if trackCount == 0 {
-		s.waitForInFlightSubscriptions()
-		s.initialized.Break()
+		s.completeInit()
 		return 0, 0, nil
 	}
 
@@ -465,8 +460,7 @@ func (s *SDKSource) awaitTracks(expecting map[string]struct{}) (uint32, uint32, 
 		}
 	}
 
-	s.waitForInFlightSubscriptions()
-	s.initialized.Break()
+	s.completeInit()
 	return w, h, nil
 }
 
--- a/pkg/pipeline/source/track_worker.go
+++ b/pkg/pipeline/source/track_worker.go
@@ -23,9 +23,9 @@ import (
 	"github.com/pion/webrtc/v4"
 	"go.uber.org/atomic"
 
+	"github.com/frostbyte73/core"
 	"github.com/livekit/egress/pkg/config"
 	"github.com/livekit/egress/pkg/errors"
-	"github.com/frostbyte73/core"
 	"github.com/livekit/egress/pkg/gstreamer"
 	"github.com/livekit/egress/pkg/pipeline/source/sdk"
 	"github.com/livekit/egress/pkg/pipeline/tempo"
@@ -361,7 +361,7 @@ func (s *SDKSource) processActiveOp(_ *trackWorker, trackID string, state *worke
 	switch op.Type {
 	case OpSubscribe:
 		// Not possible, double subscribe shouldn't be possible, nothing to do
-		logger.Warnw("unexpected Subscribe in ACTIVE state", nil, "trackID", trackID)
+		logger.Warnw("unexpected subscribe in ACTIVE state", nil, "trackID", trackID)
 
 	case OpPlaying:
 		if op.Generation == state.generation && state.writer != nil {


pkg/config/pipeline.go
```diff
@@ -84,7 +84,6 @@ type SDKSourceParams struct {
 	TrackSource  string
 	TrackKind    string
 	ScreenShare  bool
-	AudioInCodec types.MimeType
```
milos-lk, line 87: this was dead code

pkg/pipeline/source/track_worker.go
```diff
@@ -0,0 +1,505 @@
...
+	case types.MimeTypeOpus, types.MimeTypePCMU, types.MimeTypePCMA:
+		s.AudioEnabled = true
+		if s.AudioOutCodec == "" {
+			if ts.MimeType == types.MimeTypePCMU || ts.MimeType == types.MimeTypePCMA {
+				s.AudioOutCodec = types.MimeTypeOpus
+			} else {
+				s.AudioOutCodec = ts.MimeType
+			}
+		}
+		s.AudioTranscoding = true
```
boks1971, line 218: Is this always `true` even if `s.AudioOutCodec = ts.MimeType`?

pkg/pipeline/source/track_worker.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	if s.closing.Load() {
+		// Release subLock before blocking drain
+		s.subLock.RUnlock()
+		s.handleOrphanedWriter(trackID, writer)
+		return nil
+	}
+
+	s.mu.Lock()
```
boks1971, line 299: I am sure it is okay, but I don't have full view/understanding yet, so just checking, is acquiring what looks like top level lock okay to acquire after `subLock`? Is the ordering okay?

pkg/pipeline/source/track_worker.go
```diff
@@ -0,0 +1,505 @@
...
+	case TrackStateIdle:
+		return s.processIdleOp(w, trackID, ws, op)
+	case TrackStateActive:
+		return s.processActiveOp(w, trackID, ws, op)
+	case TrackStateCleaning:
+		// Unreachable: worker blocks in startCleanup while state is CLEANING.
+		// Ops queue in opChan and are processed after cleanup completes (in IDLE state).
+		return false
+	default:
+		return false
```
boks1971, line 334: can we log an explicit invalid combination of state + op so that we know if it happens and under what conditions it happens?

boks1971: I see log below for IDLE op invalid state, but I think it might be useful here also to see what combination falls into this `default` case. Maybe it is not that useful.

pkg/pipeline/source/track_worker.go
milos-lk: mixer requires decoded pcm samples in any case

pkg/pipeline/source/track_worker.go
milos-lk: yes - it's fine - lock order is always same - although we need to be careful - that's the order we kept also so far on the main branch (no change wrt lock ordering here).

milos-lk:
> What did you use to draw those diagrams?

asked claude to generate mermaid text for these graphs and inserted it with ```mermaid block in the description.

pkg/pipeline/source/track_worker.go
milos-lk: I think it doesn't hurt to have logging also for invalid/unexpected actions which don't trigger state changes - will add it

milos-lk: added

boks1971: Thank you

pkg/pipeline/source/sdk.go
```diff
@@ -261,6 +292,44 @@ func (s *SDKSource) joinRoom() error {
...
+
+// StopAwaitingTracks - called after init complete or timeout
+func (s *SDKSource) stopAwaitingTracks() {
+	s.initResultChan.Store(nil) // just nil out, don't close
+}
+
+// waitForInFlightSubscriptions blocks until all workers release their RLock.
+// This ensures in-flight subscription processing completes before initialized.Break(),
+func (s *SDKSource) waitForInFlightSubscriptions() {
+	s.subLock.Lock()
```
biglittlebigben, line 306: How are you sure that no worker will attempt to take the RLock after Unlock returns? 

Can you comment further on what the barrier protects against? It's a pretty unusual pattern and I think you need both to convince folks that it's safe, and be very clear about explaining it for future maintenance.

pkg/pipeline/source/track_worker.go
```diff
@@ -0,0 +1,508 @@
...
+	writer     *sdk.AppWriter
+	generation uint64
+}
+
+// trackWorker manages the lifecycle of a single track.
+// Each track gets its own goroutine to serialize operations and avoid cross-track blocking.
+type trackWorker struct {
+	trackID    string
+	opChan     chan Operation // buffered channel for operations
+	done       chan struct{}  // closed when worker exits
```
biglittlebigben: Should this be a core.Fuse for consistency?

pkg/pipeline/source/track_worker.go
```diff
@@ -0,0 +1,508 @@
...
+	generation atomic.Uint64  // current generation (for Playing coordination)
+}
+
+func (s *SDKSource) getOrCreateWorker(trackID string) *trackWorker {
+	// Fast path - worker exists
+	s.workersMu.RLock()
+	w, exists := s.workers[trackID]
```
biglittlebigben, line 121: Is using a RWLock critical for performance here? How can this block?

milos-lk: it's just part of double checking lock pattern to make read operations (more frequent) lighter and minimize contention on reads.

biglittlebigben: Right, we are waiting around on a channel while holding the read lock above.

pkg/pipeline/source/sdk.go
milos-lk: The barrier ensured subscriptions that started before the barrier completed their work. But it left a race window between Unlock() and Break() where new subscriptions could:
1. Get RLock (no longer blocked)
2. Capture "not initialized"
3. Then Break() runs
4. Subscription continues with wrong path

I wasn't sure what was the original intent of this lock - but after looking into it once more - I figured that `subLock` is needed to prevent the race condition when breaking initialized fuse - if a subscription sneaks in just before the fuse is broken and the fuse gets broken immediately after that - onTrackAdded won't be called (from the subscription point of view it's still and the await phase and skips onTrackAdded callback. That's now addressed with the `completeInit` method that breaks the fuse inside the critical section - so no more ugly barrier logic is needed.

milos-lk: also updated the comment for the lock to make it more obvious going forward

biglittlebigben: Ok, code makes a lot more sense to me now 😊

PR #1130, opened 2026-02-16
Potential fix for rare issue causing tracks not to be recorded

milos-lk:
Very small number of recordings are impacted by an issue where a track can't be recorded - all packets for the affected track are received but none of them can be pushed due to an unexpected flow return error from the gst pipeline. After looking more into the code - this can happen because we use appsrc state as a guard for pushing packets to pipeline - and waiting for it to be playing. However that's not the best guard - within the src bin this appsrc belongs to, other elements might not be in ready/playing state when appsrc becomes. So when pushing starts it can hit an element not yet in ready state and streaming task will deactivate. The most secure solution would be tracking src bin states instead - but since they could also be async the logic to handle that isn't super trivial - so here I opted to try another solution - using need-data signal from appsrc as an additional guard that might be just enough for making sure the whole chain of elements is properly established.

### DIFF
--- a/pkg/pipeline/source/sdk/appwriter.go
+++ b/pkg/pipeline/source/sdk/appwriter.go
@@ -100,6 +100,7 @@ type AppWriter struct {
 	lastPushed               atomic.Time
 	playing                  core.Fuse
 	srcNeedsData             core.Fuse
+	addedToPipeline          core.Fuse
 	draining                 core.Fuse
 	unsubscribed             core.Fuse
 	endStreamSignaled        core.Fuse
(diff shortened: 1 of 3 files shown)


paulwe (approved):

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -151,6 +152,14 @@ func NewAppWriter(
 		timeProvider:      gstreamer.NopTimeProvider(),
 	}
 	w.samplesCond = sync.NewCond(&w.samplesLock)
+	w.src.SetCallbacks(&app.SourceCallbacks{
```
biglittlebigben, line 156: How confident are we that this will always fire? If not totally sure, should we have an option to enable/disable this behavior?

biglittlebigben (approved):

pkg/pipeline/source/sdk/appwriter.go
milos-lk: It will aways fire: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-base/gst-libs/gst/app/gstappsrc.c?ref_type=heads#L1829

Upon starting the streaming thread - the queue inside appsrc is always empty and the else branch linked above will aways execute.

PR #1138, opened 2026-02-25
Egress auto retry [3/7]

milos-lk:
Support egress retry with output file naming and configurable failure injection

Please note - this doesn't implement auto retrying - just adds support so that API consumer could make retry decision for the same egressID and implement recording retries.

Notable changes:
- Add support for retrying failed egresses with proper output file naming.  A {retry} template placeholder is needed for files and segments to allow retries and prevent overwriting files with explicitly set paths (without retry)
- Propagate RetryCount field from StartEgressRequest into EgressInfo so the pipeline is aware of the retry count.
- Add configurable failure injection for e2e testing: when test_failure_injection is enabled in config and the room name contains _inject_failure, the first attempt fails immediately while retries succeed normally.

Depends on: https://github.com/livekit/protocol/pull/1427

### DIFF
--- a/pkg/config/output_segment.go
+++ b/pkg/config/output_segment.go
@@ -60,10 +60,14 @@ func (p *PipelineConfig) getSegmentConfig(segments *livekit.SegmentedFileOutput)
 	prefix := clean(segments.FilenamePrefix)
 	playlist := clean(segments.PlaylistName)
 
-	// On retry, explicit paths must include {retry} to avoid overwriting previous attempt.
-	// When both prefix and playlist are empty, they're auto-generated with retry count.
-	if p.Info.RetryCount > 0 && (prefix != "" || playlist != "") {
-		if !strings.Contains(prefix, "{retry}") && !strings.Contains(playlist, "{retry}") {
+	// On retry, segment filenames are "{prefix}_{index}.ts" so prefix must contain {retry}
+	// to avoid overwriting. When prefix is empty it derives from playlist name, so playlist
+	// must contain {retry}. When both are empty, names are auto-generated with retry count.
+	if p.Info.RetryCount > 0 {
+		if prefix != "" && !strings.Contains(prefix, "{retry}") {
+			return nil, errors.ErrNonRetryableOutput
+		}
+		if prefix == "" && playlist != "" && !strings.Contains(playlist, "{retry}") {
 			return nil, errors.ErrNonRetryableOutput
 		}
 	}
--- a/pkg/config/retry_test.go
+++ b/pkg/config/retry_test.go
@@ -131,11 +131,11 @@ func TestSegmentOutputRetrySafety(t *testing.T) {
 			expectErr:  false,
 		},
 		{
-			name:       "retry with {retry} in playlist only",
+			name:       "retry with {retry} in playlist only (prefix explicit)",
 			retryCount: 1,
 			prefix:     "segments/my-stream",
 			playlist:   "segments/playlist-{retry}",
-			expectErr:  false,
+			expectErr:  true,
 		},
 		{
 			name:       "retry with {retry} in both",
@@ -150,6 +150,13 @@ func TestSegmentOutputRetrySafety(t *testing.T) {
 			prefix:     "segments/my-stream",
 			playlist:   "",
 			expectErr:  true,
+		},
+		{
+			name:       "retry with {retry} in playlist only (prefix empty, derives from playlist)",
+			retryCount: 1,
+			prefix:     "",
+			playlist:   "segments/playlist-{retry}",
+			expectErr:  false,
 		},
 		{
 			name:       "retry with explicit playlist missing {retry}",


pkg/handler/handler.go
```diff
@@ -103,6 +105,17 @@ func (h *Handler) Run() {
 	var err error
 	egressID := h.conf.Info.EgressId
 
+	if h.shouldInjectEgressFailure() {
```
milos-lk, line 108: adding the special hook that could be used from tests to easily simulate egress failures.

pkg/config/base.go
```diff
@@ -65,6 +65,7 @@ type BaseConfig struct {
 	Latency              LatencyConfig                       `yaml:"latency"`                // gstreamer latencies, modifying these may break the service
 	LatencyOverrides     map[types.RequestType]LatencyConfig `yaml:"latency_overrides"`      // latency overrides for different request types, experimental only, will be removed
 	AudioTempoController AudioTempoController                `yaml:"audio_tempo_controller"` // audio tempo controller
+	TestFailureInjection bool                                `yaml:"test_failure_injection"` // staging-only: enable e2e failure injection via room name
```
biglittlebigben: It feels confusing to add this to the main OSS config as it is only relevant to our cloud operations. I can't really think of a better option right now though. Maybe at least make the option the name of the room to fail, rather than just a boolean?

pkg/config/output_file.go
```diff
@@ -134,6 +145,9 @@ func (o *FileConfig) updateFilepath(p *PipelineConfig, identifier string, replac
 				o.StorageFilepath = o.StorageFilepath[:extIdx]
 			}
 		}
+		if p.Info.Retries > 0 && !hasRetryTemplate {
```
biglittlebigben: It's unclear to me whether adding the retry suffix is desirable in the case where the customer provided the filename. We are writing files they did not ask us to write to their blobstore. This may break some of their app logic.

milos-lk: true - although filename shouldn't be used for alone for determining retry eligibility (e.g stream output only) - in case it's provided (file/hls output) the safest way would be to prevent any attempt and not hit this logic if the retry placeholder isn't there. This however could imply that service controlling and orchestrating egresses  (including retries) needs to be able to understand its configuration (and not engage retrying in case  a file input is missing the retry placeholder in the filename). On the other side egress auto retries could be made explicit opt-in via project settings - where we could document the behavior. In both case a form of explicit opt-in might be needed - not really sure what's better long term.

biglittlebigben: I really don't think that requiring opt in for retries make a lot of sense generally speaking, beyond the roll out phase. The orchestrator can send the retry request to egress, and egress can fail with an appropriate error if the request is invalid for retries.

pkg/config/base.go
milos-lk: right now it's the combo of bool and specific room name - but we can make it just TestRoomFailureInjection string and match that if set. will change.

milos-lk: updated

pkg/config/output_file.go
```diff
@@ -133,9 +130,6 @@ func (o *FileConfig) updateFilepath(p *PipelineConfig, identifier string, replac
 	if o.StorageFilepath == "" || strings.HasSuffix(o.StorageFilepath, "/") {
 		// generate filepath
 		baseName := fmt.Sprintf("%s-%s", identifier, time.Now().Format("2006-01-02T150405"))
```
biglittlebigben, line 141: I think it's fine to append the retry suffix in this case as the customer didn't provide an explicit filename

milos-lk: updated - added for image and hls files

biglittlebigben: Do we fail the retry anywhere if the filename is provided explicitly and there is no `{retry}` token in the filename?

pkg/config/output_segment.go
```diff
@@ -137,6 +137,9 @@ func (o *SegmentConfig) updatePrefixAndPlaylist(p *PipelineConfig) error {
 			playlistName = segmentPrefix
 		} else {
 			playlistName = fmt.Sprintf("%s-%s", identifier, time.Now().Format("2006-01-02T150405"))
+			if p.Info.RetryCount > 0 {
+				playlistName = fmt.Sprintf("%s-%d", playlistName, p.Info.RetryCount)
```
biglittlebigben, line 156: How about segment names? Is there anything guarantee that the segment names will be different across retries without a suffix?

pkg/config/base.go
```diff
@@ -65,6 +65,7 @@ type BaseConfig struct {
 	Latency              LatencyConfig                       `yaml:"latency"`                // gstreamer latencies, modifying these may break the service
 	LatencyOverrides     map[types.RequestType]LatencyConfig `yaml:"latency_overrides"`      // latency overrides for different request types, experimental only, will be removed
 	AudioTempoController AudioTempoController                `yaml:"audio_tempo_controller"` // audio tempo controller
+	TestFailureInjectionRoom string                              `yaml:"test_failure_injection_room"` // staging-only: inject failure for rooms containing this substring
```
biglittlebigben: Should we move this to a "TestOverrides" structure defined in a different file maybe, just for clarity?

pkg/config/output_file.go
milos-lk: at the moment no - most of the issues we have now are around starting a recording - doing this opportunistically will cover all of these with no/minimal loss. Also for cases where a failure happens mid recording - then we will just have the second part of the recording.. doesn't make the situation worse for file outputs (since in either case that part is lost) though - but it will break HLS (if we use the same name) - just realizing that. wdyt - should we allow for file recordings to re-use the same name and still retry (original file not uploaded) - and fail only HLS streams? If we don't want to retry we need to do one of these 2:
- return something specific in error response that would disallow retries
- make orchestrating service aware of the egress config part so that it knows it shouldn't retry

pkg/config/base.go
milos-lk: sure - won't hurt. will update this

pkg/config/output_file.go
biglittlebigben: I think we need to be very conservative here. Overwriting a key in the customer's bucket is never OK IMHO. We should fail the retry, returning the error as well as an indication of why we didn't retry if we can't generate a different filename.

milos-lk: I agree - the point about file still holds though - if egress failed there is nothing to overwrite.

biglittlebigben: This brings back another topic we discussed in the past: I think we need to have different indexes for the retry index (that we use for observability and logging) and filename index, that we increase only after the session status reached to "started".

milos-lk: I am looking into a couple of options now - and not finding a solution I like - all options I got so far have something ugly - it might be worth having a short chat again on this topic.

milos-lk: updated as discussed - for files/segments we are going to refuse retry requests if the path is explicitly specified without {{retry}} param. Added some config unit tests for that.

pkg/config/base.go
milos-lk: updated

pkg/config/output_segment.go
milos-lk: I thought there is - but there is a subtle case I didn't cover - if both prefix and playlist are set {retry} inside playlist name doesn't matter as it won't become part of the segment name - and in case of retries they will collide - so great catch - pushed the fix, updated / added test cases

PR #1142, opened 2026-03-01
Heal track which enters into continuous flow flushing loop on pushing packets

milos-lk:
When this rare issue occurs - instead of just continuously pushing data for the affected track and getting the error and ending up with EOS not being forwarded for it (which leads to pipeline frozen errors) - remove the affected appsrc from the pipeline and add it again. The issue seems to happen in a narrow window of adding a track src bin when pipeline itself is transitioning to PLAYING - so this change would make retry removing and adding source bin again outside of this window - only after pipeline is in playing state. Thus it would have a fair chance of succeeding and "healing the track".

The issue is impossible to reproduce locally - so to test this I did sort of mutation testing - changed the code to simulate the error for one of tracks in the recording and removing/adding the appsrc worked as expected.

### DIFF
--- a/pkg/gstreamer/bin.go
+++ b/pkg/gstreamer/bin.go
@@ -167,7 +167,7 @@ func (b *Bin) AddElements(elements ...*gst.Element) error {
 // This is used for FlowFlushing recovery where EOS will never propagate from a stuck appsrc.
 // The removal runs on the GLib main loop thread via glib.IdleAdd and blocks until complete.
 func (b *Bin) ForceRemoveSourceBin(name string) error {
-	logger.Debugw(fmt.Sprintf("force removing src %s from %s", name, b.bin.GetName()))
+	logger.Infow("force removing source bin", "src", name, "from", b.bin.GetName())
 
 	b.LockStateShared()
 	defer b.UnlockStateShared()


pkg/gstreamer/bin.go
```diff
@@ -162,6 +163,61 @@ func (b *Bin) AddElements(elements ...*gst.Element) error {
 	return nil
 }
 
+// ForceRemoveSourceBin synchronously removes a source bin without waiting for EOS.
+// This is used for FlowFlushing recovery where EOS will never propagate from a stuck appsrc.
+// The removal runs on the GLib main loop thread via glib.IdleAdd and blocks until complete.
+func (b *Bin) ForceRemoveSourceBin(name string) error {
+	logger.Debugw(fmt.Sprintf("force removing src %s from %s", name, b.bin.GetName()))
```
boks1971: Use longing fields so that filtering is easier?

Also should this be Infow?

pkg/gstreamer/bin.go
```diff
@@ -162,6 +163,61 @@ func (b *Bin) AddElements(elements ...*gst.Element) error {
...
+	done := make(chan error, 1)
+	if _, err := glib.IdleAdd(func() bool {
+		logger.Debugw("force removing source bin on GLib thread", "bin", src.bin.GetName())
+		done <- detachSourceBin(src, srcGhostPad, sinkGhostPad, peerElement, parentBin, pipeline)
+		return false
+	}); err != nil {
+		return errors.ErrGstPipelineError(err)
+	}
+
+	return <-done
```
boks1971, line 218: Don't understand this function, but looks a bit scary. There locks and waiting on a channel. I don't have any suggestions, but just noting it.

boks1971 (approved):

pkg/gstreamer/bin.go
milos-lk: it's fair concern - the reason we block here on done is that caller needs to know removal finished before proceeding with adding a new bin. It's very similar logic to what existing probeRemoveSource does (just here we can't afford fire and forget approach - I extracted the common part to detachSourceBin() function).
It should be fine since:
- we release b.mu before waiting on done, the only lock held is `LockStateShared()` - RLock - and nothing inside callback tries any operation on it (as the callback logic is low level and doesn't involve management of that state).
- callback always returns

Not sure if I can come up with something simpler.

pkg/gstreamer/bin.go
```diff
@@ -162,6 +163,61 @@ func (b *Bin) AddElements(elements ...*gst.Element) error {
...
+
+	b.LockStateShared()
+	defer b.UnlockStateShared()
+
+	state := b.GetStateLocked()
+	if state > StateRunning {
+		return nil
+	}
+
+	b.mu.Lock()
```
biglittlebigben, line 180: 2 locks are held from this point on...

pkg/gstreamer/bin.go
```diff
@@ -162,6 +163,61 @@ func (b *Bin) AddElements(elements ...*gst.Element) error {
...
+
+	idx := slices.IndexFunc(b.srcs, func(s *Bin) bool { return s.bin.GetName() == name })
+	if idx == -1 {
+		b.mu.Unlock()
+		return nil
+	}
+	src := b.srcs[idx]
+
+	src.mu.Lock()
+	srcGhostPad, sinkGhostPad, ok := deleteGhostPadsLocked(src, b)
```
biglittlebigben, line 190: reaching into the guts of a different object to catch a lock is a bit iffy. Should we call a function that takes the lock itself?

pkg/gstreamer/bin.go
biglittlebigben: Is there any way to safely add a watchdog here? Similar to what "Pipeline Frozen" is in practice...

pkg/gstreamer/bin.go
milos-lk: it's exactly the same pattern locking and accessing wise as we have inside the existing remove bin. There are 2 different aspects of this API - it's made for sources only and it needs to perform operation synchronously without waiting for EOS. I am not saying it's particularly nice - if we want to change/simplify things here - we would need to rethink the logic inside bin.go entirely. Didn't want to open that particular can of worms here.
Having 2 locks isn't that scary though (although it requires being specially careful when modifying this code) - since they are taken in the same order as abstraction layers - before making the change I was particularly analyzing the removeBin function for correctness wrt concurrency and couldn't find issues tbh - is there any specific use case that you saw could cause the issue here?

pkg/gstreamer/bin.go
milos-lk: we probably should - this is taken from the existing removeBin - I will take a look and see if we can encapsulate this 👍

pkg/gstreamer/bin.go
milos-lk: I don't think it's worth it tbh - for 2 reasons:
- even in the theoretical case where nothing comes out of that channel, gortoutine performing pushing packets will be blocked on this and won't be able to send to EOS at the end, at which point the existing pipeline frozen check engages marking egress failed and killing the handler anyways (we have the same behavior today even without this change) 
- forcing src bin removal should only be performed in very rare occasions when graceful pipeline manipulation is impossible - having a watch dog on top of the caller of this call stack could theoretically allow us to fail faster (although that should be implemented as well) - but I thought increasing complexity to cover edge case (which I tbh don't see happening) of an edge case for having early returns might be excessive

Do you think we have some cases where this could be beneficial?

pkg/gstreamer/bin.go
milos-lk: for the time being it's worth maybe documenting lock ordering explicitly here so it's a bit easier for making future changes to this code base - whoever needs to change something could rely on that instead of re-building the context.

biglittlebigben: 👍

PR #1143, opened 2026-03-03
Documenting locking rules for bin.go

milos-lk:
As per comment here:
https://github.com/livekit/egress/pull/1142/changes#r2874819315

### DIFF
--- a/pkg/gstreamer/bin.go
+++ b/pkg/gstreamer/bin.go
@@ -40,8 +40,15 @@ const (
 //     bin mutexes (for example: b.mu -> src.mu -> sink.mu).
 //  3. do not introduce paths that acquire locks in the reverse order
 //     (peer/child -> parent), or AB-BA deadlocks are possible.
-//  4. for work executed later on the GLib loop (IdleAdd callbacks), snapshot any
-//     shared fields while holding the lock, then unlock before scheduling work.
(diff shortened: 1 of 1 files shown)


pkg/gstreamer/bin.go
```diff
@@ -33,6 +33,16 @@ const (
...
 
+// Locking rules for Bin/StateManager (maintainer reference):
+//  1. if both state and bin data are needed, take StateManager lock first
+//     (LockState/LockStateShared), then Bin.mu.
+//  2. for multi-bin operations, take the "owner"/parent bin mutex before peer
+//     bin mutexes (for example: b.mu -> src.mu -> sink.mu).
+//  3. do not introduce paths that acquire locks in the reverse order
+//     (peer/child -> parent), or AB-BA deadlocks are possible.
+//  4. for work executed later on the GLib loop (IdleAdd callbacks), snapshot any
+//     shared fields while holding the lock, then unlock before scheduling work.
```
boks1971: which lock is this? Would be good to explicitly mention which lock.

And maybe also explicitly mention that holding state lock across GLib loop?

milos-lk: good point - updated

milos-lk:
> Thanks. Real question is: will Claude understand this...

I bet it doesn't need it :)

PR #1148, opened 2026-03-14
Update server sdk to include fix for reported file duration inflation

milos-lk:
This bumps min go version to 1.25 and with it comes required updating golangci linters to v2 - configuration has changed and some more warnings are now generated, added them to the exclusion list (to be addressed independently as these aren't new).

### DIFF
--- a/.github/workflows/publish-egress.yaml
+++ b/.github/workflows/publish-egress.yaml
@@ -23,7 +23,7 @@ on:
       - 'v*.*.*'
 jobs:
   docker:
-    runs-on: buildjet-8vcpu-ubuntu-2204
+    runs-on: namespace-profile-8vcpu-cache
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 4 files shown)


.golangci.yaml
```diff
@@ -1,66 +1,72 @@
+version: "2"
```
milos-lk, line 1: new version got autogenerated based on the old one by running a migration script

.golangci.yaml
```diff
@@ -1,66 +1,72 @@
...
+      forbid:
+        - pattern: sync\.Mutex
+        - pattern: sync\.RWMutex
+      analyze-types: true
+    staticcheck:
+      checks:
+        - "all"
+        - "-QF*"  # quick-fix suggestions — too noisy, address separately
+        - "-S*"   # simplification suggestions — too noisy, address separately
+        - "-ST*"  # style checks — too noisy, address separately
```
milos-lk, line 31: these would be nice to address but would need to change a lot of independent files now. Will do this separately.

boks1971 (approved):

PR #1155, opened 2026-03-20
v2 api

frostbyte73:
https://www.notion.so/livekit/Egress-v2-Replay-Export-6813d87e17e8437db6fd8ee759df7b13?source=copy_link#32e3c4901a4280ccb693cb8d73268f01

### DIFF
--- a/pkg/pipeline/source/sdk.go
+++ b/pkg/pipeline/source/sdk.go
@@ -99,7 +99,7 @@ func NewSDKSource(ctx context.Context, p *config.PipelineConfig, callbacks *gstr
 		}),
 	}
 
-	if p.RequestType == types.RequestTypeRoomComposite {
+	if p.RequestType == types.RequestTypeRoomComposite || p.RequestType == types.RequestTypeTemplate {
 		// Enable Packet Burst Estimator for Room Composite requests
 		opts = append(opts, synchronizer.WithStartGate())
 	}
@@ -235,17 +235,12 @@ func (s *SDKSource) joinRoom() error {
 		OnDisconnected: s.onDisconnected,
 	}
 
-	if s.RequestType == types.RequestTypeRoomComposite {
+	switch s.RequestType {
+	case types.RequestTypeRoomComposite, types.RequestTypeTemplate, types.RequestTypeMedia:
 		cb.OnTrackPublished = s.onTrackPublished
-	}
-
-	if s.RequestType == types.RequestTypeParticipant {
+	case types.RequestTypeParticipant:
 		cb.OnTrackPublished = s.onTrackPublished
 		cb.OnParticipantDisconnected = s.onParticipantDisconnected
-	}
-
-	if s.RequestType == types.RequestTypeMedia {
-		cb.OnTrackPublished = s.onTrackPublished
 	}
 
 	logger.Debugw("connecting to room")
@@ -261,7 +256,11 @@ func (s *SDKSource) joinRoom() error {
 	case types.RequestTypeRoomComposite:
 		fileIdentifier = s.room.Name()
 		// room_name and room_id are already handled as replacements
+		err = s.awaitRoomTracks()
 
+	case types.RequestTypeTemplate:
+		fileIdentifier = s.room.Name()
+		s.filenameReplacements["{room_name}"] = s.room.Name()
 		err = s.awaitRoomTracks()
 
 	case types.RequestTypeParticipant:
@@ -286,6 +285,7 @@ func (s *SDKSource) joinRoom() error {
 
 	case types.RequestTypeMedia:
 		fileIdentifier = s.room.Name()
+		s.filenameReplacements["{room_name}"] = s.room.Name()
 		w, h, err = s.awaitMediaTracks()
 	}
 	if err != nil {
@@ -627,7 +627,10 @@ func (s *SDKSource) onTrackSubscribed(track *webrtc.TrackRemote, pub *lksdk.Remo
 }
 
 func (s *SDKSource) onTrackPublished(pub *lksdk.RemoteTrackPublication, rp *lksdk.RemoteParticipant) {
-	if s.RequestType != types.RequestTypeParticipant && s.RequestType != types.RequestTypeRoomComposite && s.RequestType != types.RequestTypeMedia {
+	if s.RequestType != types.RequestTypeParticipant &&
+		s.RequestType != types.RequestTypeRoomComposite &&
+		s.RequestType != types.RequestTypeTemplate &&
+		s.RequestType != types.RequestTypeMedia {
 		return
 	}
 
@@ -660,7 +663,7 @@ func (s *SDKSource) shouldSubscribe(pub lksdk.TrackPublication) bool {
 		default:
 			return s.ScreenShare
 		}
-	case types.RequestTypeRoomComposite:
+	case types.RequestTypeRoomComposite, types.RequestTypeTemplate:
 		switch pub.Kind() {
 		case lksdk.TrackKindAudio:
 			return s.AudioEnabled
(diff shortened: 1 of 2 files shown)


pkg/config/output.go
```diff
@@ -217,6 +217,188 @@ func (p *PipelineConfig) updateEncodedOutputs(req egress.EncodedOutput) error {
 	return nil
 }
 
+func (p *PipelineConfig) updateOutputs(req *livekit.ExportReplayRequest) error {
```
milos-lk, line 220: updateOutputs (lines 131-311) is a near-copy of updateEncodedOutputs.
Do we plan to get rid of the latter? If not (maybe in any case)  I would suggest extracting output processing helpers rather than duplicating good chunk of the existing function. e.g:
- processFileOutput (body inside OutputFile case
- processStreamOutput
- processSegmentOutput
- processImageOutput

Both updateEncodedOutputs and updateOutputs would call these helpers differing only in how they extract  parameters from their respective request types.

pkg/config/pipeline.go
```diff
@@ -577,6 +725,22 @@ func (p *PipelineConfig) updateOutputType(compatibleAudioCodecs map[types.MimeTy
 	return nil
 }
 
+// TemplateUsesSDKSource reports whether a template source request will use
+// the SDK source (no Chrome/Pulse) instead of Web.
+// Same heuristic as RoomCompositeUsesSDKSource.
```
milos-lk, line 730: do you think it's worth reducing repetition by having something like:
```
  type sdkSourceCompatibilityChecker interface {
      GetLayout() string           
      GetAudioOnly() bool                                                                                           
      GetCustomBaseUrl() string
  }                                                                                                                 
                  
  func UsesSDKSource(req sdkSourceCompatibilityChecker) bool {
      return req.GetLayout() == "" && req.GetAudioOnly() && req.GetCustomBaseUrl() == ""                            
  }                                                                                     
``` 
Both proto types already have the generated Get methods, so this works with zero changes to the proto. Then`RoomCompositeUsesSDKSource` and `TemplateUsesSDKSource` both become UsesSDKSource(req)?

pkg/config/pipeline.go
```diff
@@ -396,6 +410,144 @@ func (p *PipelineConfig) Update(request *rpc.StartEgressRequest) error {
...
+			case *livekit.MediaSource_VideoTrackId:
+				p.VideoEnabled = true
+				p.VideoDecoding = true
+				p.VideoTrackID = v.VideoTrackId
+			case *livekit.MediaSource_ParticipantVideo:
+				p.VideoEnabled = true
+				p.VideoDecoding = true
+				p.Identity = v.ParticipantVideo.Identity
+				p.ScreenShare = v.ParticipantVideo.PreferScreenShare
+			}
```
milos-lk, line 505: do you think it's worth adding a case that would catch replay request for multiple video tracks and media source and error out as not supported at the moment?

pkg/config/output.go
frostbyte73: I plan to convert the requests and then entirely remove the old paths

pkg/config/pipeline.go
frostbyte73: right now it's not possible, if they did specify a video track and a video participant it would only use the specified track

pkg/pipeline/source/sdk.go
```diff
@@ -463,6 +473,83 @@ func (s *SDKSource) awaitTracks(expecting map[string]struct{}) (uint32, uint32,
 	return w, h, nil
 }
 
+func (s *SDKSource) awaitMediaTracks() (uint32, uint32, error) {
```
milos-lk, line 476: this should replace existing await functions in future right?

pkg/pipeline/source/sdk.go
```diff
@@ -463,6 +473,83 @@ func (s *SDKSource) awaitTracks(expecting map[string]struct{}) (uint32, uint32,
...
+	// the audioChannels map before subscribing.
+	var expected, alreadySubscribed int
+	for _, rp := range s.room.GetRemoteParticipants() {
+		for _, pub := range rp.TrackPublications() {
+			if s.shouldSubscribeMedia(pub, rp) {
+				if pub.IsSubscribed() {
+					alreadySubscribed++
+					continue // Already subscribed via onTrackPublished
+				}
+				if err := s.subscribe(pub); err != nil {
```
milos-lk, line 490: is this a race possibility - we are subscribing to the track before the result channel is installed. If the subscription is fast enough and finishes the round trip (and gets onSubscribed callback) before we get to the  `startAwaitingTracks` line - the result channel will be nil and the startup will silently miss both success or failure?

pkg/pipeline/source/sdk.go
```diff
@@ -463,6 +473,83 @@ func (s *SDKSource) awaitTracks(expecting map[string]struct{}) (uint32, uint32,
...
+			if pub.IsSubscribed() && pub.Kind() == lksdk.TrackKindVideo {
+				if info := pub.TrackInfo(); info != nil {
+					w = info.Width
+					h = info.Height
+				}
+			}
+		}
+	}
+
+	s.completeInit()
```
milos-lk, line 549: What's the reason awaitMediaTracks() doesn't wait for all currently matched initial subscriptions here (or the 3s soft timeout)? awaitExpected() keeps consuming results until it reaches the expected count, but this path completes init after a single receive, which seems to let startup proceed while other requested media tracks are still in flight. That would make later successes post-init dynamic adds and later failures non-startup-fatal. Looks like behavior change.

pkg/config/pipeline.go
```diff
@@ -409,10 +561,6 @@ func (p *PipelineConfig) Update(request *rpc.StartEgressRequest) error {
 
 	// connection info
 	if connectionInfoRequired {
-		if p.Info.RoomName == "" {
```
milos-lk, line 412: I am assuming this is done for replay - legacy requests with an explicit token but no room_name will now pass validation and run with Info.RoomName == "", which changes observable behavior in filename expansion (output_file) and manifest generation. Or I missed place where we would reject these?

pkg/info/io.go
```diff
@@ -265,7 +265,11 @@ func (c *sessionReporter) handleUpdate(w *worker, egressID string) {
 		if !c.setHealthy(true) {
 			logger.Infow("io connection restored", "egressID", u.info.EgressId)
 		}
-		requestType, outputType := egress.GetTypes(u.info.Request)
+		var typesInput interface{} = u.info.Request
```
milos-lk, line 268: nit: any

pkg/pipeline/builder/audio.go
```diff
@@ -400,24 +396,31 @@ func (b *AudioBin) resetAudioAppSrcBin(ts *config.TrackSource) error {
...
+			return livekit.AudioChannel_AUDIO_CHANNEL_LEFT
 		}
-		return audioChannelRight
+		return livekit.AudioChannel_AUDIO_CHANNEL_RIGHT
 
 	case livekit.AudioMixing_DUAL_CHANNEL_ALTERNATE:
-		next := b.nextChannel
-		b.nextChannel++
-		return next%2 + 1
+		if b.nextChannel == livekit.AudioChannel_AUDIO_CHANNEL_LEFT {
```
milos-lk, line 415: very simple and makes it much easier to read 👍

pkg/pipeline/source/sdk.go
frostbyte73: Yeah, same situation

pkg/config/pipeline.go
frostbyte73: Token generation fails with "no room name" when you pass an empty string, so behavior is unchanged there

PR #1185, opened 2026-04-11
delete the image, not just the tag

frostbyte73:


paulwe (approved):

frostbyte73:
Apparently deleting the tag is the correct method, they'll get gc'd later

PR #1188, opened 2026-04-14
fix HandlerLogger

frostbyte73:


paulwe (approved):

paulwe (commented):
tests?

pkg/logging/handler.go
```diff
@@ -1,63 +1,126 @@
...
 )
 
-var sdkPrefixes = map[string]bool{
-	"turnc": true, // turnc ERROR
-	"ice E": true, // ice ERROR
-	"pc ER": true, // pc ERROR
-	"twcc_": true, // twcc_sender_interceptor ERROR
-	"SDK 2": true, // SDK 2025
+type HandlerLogger struct {
+	mu       deadlock.Mutex
```
boks1971, line 14: is the the use of deadlock mutex intentional?

boks1971 (approved):

pkg/logging/handler.go
frostbyte73: yes, we've replaced sync entirely in egress

frostbyte73:
> tests?

just looking at the integration test output should be enough

pkg/logging/handler.go
```diff
@@ -1,63 +1,126 @@
...
+		}
+	}
+
+	// flush panic
+	h.flushPanic()
+	return nil
+}
+
+func (h *HandlerLogger) processLine(line string) {
+	if line[len(line)-1] == '}' {
```
milos-lk, line 69: nit: maybe it makes more sense to move len(line) check inside to prevent future usages of this API without checking len?

frostbyte73: will move in another PR

pkg/logging/handler.go
```diff
@@ -1,63 +1,126 @@
...
-				strings.HasSuffix(line, "is not mapped"),
-				strings.HasSuffix(line, "load cuda library"),
-				strings.Contains(line, "libcuda.so.1"):
-				continue
-
-			default:
-				l.Errorw(line, nil)
-			}
+func (h *HandlerLogger) Write(p []byte) (int, error) {
+	h.mu.Lock()
```
milos-lk, line 30: just double checking - lock contention here might be a concern for verbose logging? The first idea coming to my mind here is having a worker gorotine that would process the buffer and write locking scope reduces and doesn't involve io

PR #1195, opened 2026-04-22
sdk compositing

frostbyte73:
Adds sdk video compositing for room composite and template egress
- Layouts match the standard chrome templates
- VideoBin rework: input-selector -> compositor for all sdk video, even when there's only one video track. This is more reliable than the old input selector, because we just have videoTestSrc running the whole time at the bottom of the composited stack, and there's no need to swap pads/tracks due to muting or disconnections

Misc:
- Fixes image sink's multifilesink properties (could hang when expecting a preroll)
- Fixes race on rp.Kind() in sdk source path


pkg/pipeline/builder/image.go
```diff
@@ -122,6 +122,13 @@ func BuildImageBin(c *config.ImageConfig, pipeline *gstreamer.Pipeline, p *confi
 	if err != nil {
 		return nil, err
 	}
+	// async=true here would gate pipeline preroll on the first composited frame, stalling PAUSED→PLAYING.
+	if err = sink.SetProperty("sync", false); err != nil {
```
biglittlebigben, line 126: What is the behavior when both sync and asynchronous are false?

pkg/pipeline/builder/layout.go
```diff
@@ -0,0 +1,581 @@
...
+	innerH := l.height - 2*gridGap
+	availableForCols := innerW - gridGap
+	carouselW := availableForCols / totalFr
+	stageW := availableForCols - carouselW
+	stageX := gridGap + carouselW + gridGap
+
+	// thumbH always divides by maxTiles so tile size stays stable as participants join/leave.
+	maxTiles := maxCarouselTiles(carouselW, innerH)
+	thumbH := (innerH - gridGap*(maxTiles-1)) / maxTiles
+	visibleThumbs := maxTiles
```
biglittlebigben, line 302: nit: use `min`?

pkg/pipeline/builder/video.go
```diff
@@ -324,30 +390,35 @@ func (b *VideoBin) buildSDKInput() error {
 	b.pads = make(map[string]*gst.Pad)
 	b.names = make(map[string]string)
 
-	// add selector first so pads can be created
 	if b.conf.VideoDecoding {
-		if err := b.addSelector(); err != nil {
+		if err := b.addCompositor(); err != nil {
```
biglittlebigben: It looks like we are always adding the compositor if we handle video. How optimized if the GST compositor if we have no layout? Is it still going to force a conversion to RGB, or some blitting? The module definition makes me thing that the buffer will remain untouched but can anybody confirm?

biglittlebigben (commented):
Could provide a Description to the PR (and commit message) for future reference?

pkg/pipeline/builder/image.go
frostbyte73: `sync=false` - buffers are written immediately as they arrive instead of waiting for pipeline clock to match
`async=false` - state changes are performed synchronously (no prerolling)

pkg/pipeline/builder/video.go
frostbyte73: aggregator negotiates color space and supports all the major ones, but blit does always run. It does add overhead, but I think it's cleaner than the old input-switcher swapping the video track and the videotestsrc around

pkg/pipeline/builder/image.go
biglittlebigben: Some examplar API design from GST here 😊 /s

pkg/pipeline/builder/video.go
biglittlebigben: In my experience a non GPU accelerated blit is a non negligible overhead (basically a memcpy that may or may not be well optimized depending on whether they have specific code to handle basic cases). Can we measure the impact?

frostbyte73: Yeah, I'll set something up

frostbyte73: Single H264 720p track → MP4, 30s recording. ARM64 darwin docker, identical input, 5 runs each.

  | Metric             | Compositor (mean ± σ) | Input-Selector (mean ± σ) | Δ           | Δ%      |
  |--------------------|------------------------|----------------------------|-------------|---------|
  | **Total CPU**      | 14,217 ± 138 ms        | 14,015 ± 179 ms            | **+202 ms** | +1.44%  |
  | User CPU           | 13,605 ± 140 ms        | 13,455 ± 173 ms            | +150 ms     | +1.11%  |
  | Sys CPU            | 612 ± 24 ms            | 560 ± 44 ms                | +52 ms      | +9.2%   |
  | CPU as % of core   | 41.2% ± 0.4            | 40.7% ± 0.6                | **+0.49 pp**| +1.19%  |
  | Wall time          | 34,541 ± 8 ms          | 34,455 ± 75 ms             | +86 ms      | +0.25%  |
  | Max RSS            | 497 MB ± 2.6 MB        | 453 MB ± 4.3 MB            | **+45 MB**  | +9.9%   |

  _RSS excludes one outlier per condition (Go GC heap-growth blip). Raw RSS means: 534 MB vs 481 MB._

  **Per-job compositor overhead:** ~200 ms CPU (~0.6% of one core) and ~45 MB RSS, with wall time unchanged.

pkg/pipeline/builder/layout.go
```diff
@@ -0,0 +1,581 @@
...
+}
+
+func (l *LayoutManager) Layout() []PadLayout {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	return l.recalculate()
+}
+
+func (l *LayoutManager) recalculate() []PadLayout {
```
milos-lk, line 193: LayoutManager isn't aware of track being muted - wouldn't it re-apply alpha 1 on change - i.e does layout manager need to preserve track muted state, so that it can re-calculate correctly?

pkg/pipeline/builder/video.go
```diff
@@ -503,26 +592,24 @@ func (b *VideoBin) buildAppSrcBin(ts *config.TrackSource, name string) (*gstream
...
 	}
-	videoTestSrc.SetArg("pattern", "black")
 
-	queue, err := gstreamer.BuildQueue("video_test_src_queue", b.conf.Latency.PipelineLatency, false)
+	videoRate, err := gst.NewElement("videorate")
 	if err != nil {
-		return err
+		return errors.ErrGstPipelineError(err)
 	}
-	if err = queue.SetProperty("min-threshold-time", uint64(2e9)); err != nil {
```
milos-lk: Don't we need this one for non video compositing cases to ensure monotonic PTS output (the reason it was originally added for)?

milos-lk:
With EnableTemplateSDK = true hardcoded and no test setting a custom base URL, every room composite/template test now runs the SDK compositor, while the flag defaults to false in prod. Should we keep a subset running with the flag off so the production-default Chrome template path retains integration coverage?

pkg/pipeline/builder/video.go
```diff
@@ -192,14 +210,69 @@ func (b *VideoBin) onTrackUnmuted(trackID string) {
...
+	if pads != nil {
+		b.applyLayout(pads)
+	}
+}
+
+func (b *VideoBin) applyLayout(pads []PadLayout) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	b.applyLayoutLocked(pads)
```
milos-lk: applyLayoutLocked calls setVideoDimensions which calls SDKSource.UpdateTrackDimensions which results in network io while holding the mutex, maybe even when dimensions didn't change. Worth debouncing or moving outside the lock?

pkg/pipeline/builder/video.go
```diff
@@ -192,14 +210,69 @@ func (b *VideoBin) onTrackUnmuted(trackID string) {
...
+	for _, pl := range pads {
+		name, ok := b.names[pl.TrackID]
+		if !ok {
+			continue
+		}
+		pad, ok := b.pads[name]
+		if !ok {
+			continue
+		}
+		pad.SetProperty("xpos", pl.X)
```
milos-lk: nit: maybe adding a wrapper which would check the error and log e.g a warning if SetProperty fails?

pkg/pipeline/builder/video.go
```diff
@@ -501,29 +666,78 @@ func (b *VideoBin) buildAppSrcBin(ts *config.TrackSource, name string) (*gstream
...
+
+	var left, right, top, bottom int
+	if cellWiderThanSource {
+		keep := srcW * cellH / cellW
+		top = (srcH - keep) / 2
+		bottom = srcH - keep - top
+	} else {
+		keep := srcH * cellW / cellH
+		left = (srcW - keep) / 2
+		right = srcW - keep - left
```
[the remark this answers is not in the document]
frostbyte73: This one's a false positive. The `videocrop` is linked after `newVideoCapsFilter`, which pins every input to `conf.Width x conf.Height` — see `addVideoConverter` (video.go:1029) adding the chain ending in the capsfilter, then the crop appended at :675, with `Bin.AddElement` linking in insertion order (`gstreamer/bin.go:169`, `:523`). So a simulcast layer switch changes what `videoscale` receives, never what `videocrop` receives; the canvas dimensions in `setCoverCrop` are enforced upstream rather than assumed. The crop also can't exceed the frame: in each branch `keep` is strictly bounded by the corresponding source dimension.

pkg/config/base.go
```diff
@@ -41,6 +41,7 @@ type BaseConfig struct {
 	// optional
 	Logging              *logger.Config `yaml:"logging"`                // logging config
 	TemplateBase         string         `yaml:"template_base"`          // custom template base url
+	CustomTemplateBase   bool           `yaml:"custom_template_base"`   // derived at load, not operator-set
```
biglittlebigben: This probably deserves a comment. It also doesn't really belong into this structure if it is not user set...

frostbyte73: Just going to remove it, I think if you're enabling sdk templating it shouldn't care if you had a custom template url set or not

milos-lk:
I would need more time tomorrow to go through the latest changes but just briefly pointing out what I caught as potential issues until then (spent only a fraction of time I wanted looking into this so far):
- unmute might ignore layout visibility - right? Unmuting a track that the layout has hidden (e.g non-focused in single-speaker) sets its compositor pad alpha to 1.0 while W/H are still 0, so the compositor draws it at full input size behind the tiles and it could bleed through every gap until the next layout recalc?
- on main, addVideoConverter only added videorate plus a framerate capsfilter when !VideoDecoding. The PR adds them to every decoded appsrc bin, so participant and track composite egress now get a per input videorate before the input-selector, in addition to the existing one after it
- chrome room composite path loses integration coverage - EnableTemplateSDK is turned on globally so every room composite and template test now runs the SDK compositor, we should probably have subset of tests covering chrome recordings

pkg/config/pipeline.go
```diff
@@ -561,29 +568,43 @@ func ShouldUseSDKSource(req interface {
 	return req.GetAudioOnly() && req.GetLayout() == "" && req.GetCustomBaseUrl() == ""
 }
 
-func IsSDKSourceRequest(req *rpc.StartEgressRequest) bool {
+// UsesTemplateSDKCompositing reports whether a request takes the SDK compositor instead of Chrome
+func (c *BaseConfig) UsesTemplateSDKCompositing(customBaseUrl string) bool {
+	return c.EnableTemplateSDK && customBaseUrl == ""
```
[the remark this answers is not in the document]
frostbyte73: Intentional - if you have SDK templating enabled, it should use SDK templating.

milos-lk:
I think there is one remaining question worth checking with `EnableTemplateSDK = true` in the runner and no test setting a custom base URL, every room composite and template test now runs the compositor. The only Chrome template test left is the pulse sink reaper one. But prod defaults to `enable_template_sdk: false`, so the path we actually ship has no integration coverage for recordings, layouts or A/V sync anymore.

Can we keep a small subset (one per layout plus audio-only) running with the flag off? Note `test/layout.go` was rewritten to the compositor geometry, so that subset will need its own region expectations.

milos-lk:
just out of curiosity - how are we doing performance wise in case of video room composites when comparing to chrome? have you had a chance to measure it?

milos-lk:
those integration test failures have the same signature as the ones we had yesterday - I pushed the fix - so maybe it's worth merging/rebasing main

PR #1196, opened 2026-04-22
update readme for Azure SAS Token

chenosaurus:


biglittlebigben (commented):
It doesn't look like support has actually been added to egress yet. Can we wait until https://github.com/livekit/storage/pull/21 has been merged and do this update at the same time we update the livekit/storage module?

PR #1197, opened 2026-04-22
integration test overhaul

frostbyte73:
- Each participant has a timer that flashes a different color once per second, and a different audio tone that beeps once per second
- Participants mute their audio in a rotation so that only one is "talking" at a time
- Audio timestamps and video flash timestamps are captured in the output and compared


biglittlebigben (approved):
LGTM. @milos-lk may have some input as well as he made some changes here some time ago.

test/samples/participant_0_440hz.ogg
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
milos-lk, line 1: do you think media-samples would be better home for this? all testing assets we have pushed there so far. These might be useful for e2e tests - so storing them in a single place comes with that nice re-usablity effect.

test/generate_test_media.sh
```diff
@@ -0,0 +1,119 @@
...
+  RAW_VIDEO="color=c=black:s=${VIDEO_SIZE}:d=${DURATION}:r=${VIDEO_FPS}"
+
+  # Video filter: two drawtext layers for the flash effect.
+  # Grey timer shown most of the time, colored timer shown for first 90ms each second.
+  VF="drawtext=text='${TIMER_TEXT}':${TIMER_POS}:fontcolor=0x333333:enable='gte(mod(t,1),0.08)'"
+  VF="${VF},drawtext=text='${TIMER_TEXT}':${TIMER_POS}:fontcolor=0x${color_hex}:enable='lt(mod(t,1),0.08)'"
+  VF="${VF},format=yuv420p"
+
+  # Audio: brief beep (50ms) once per second, matching the video flash cadence.
+  BEEP_EXPR="sin(${freq}*2*PI*t)*(lt(mod(t,1),0.05))"
```
milos-lk, line 70: this sh reminds me of this:
https://github.com/livekit/media-samples/blob/main/test-content-generator.sh

if we move it to media samples we can extend that one and merge it with this.
When it comes to audio specifically - the existing tone generator encodes running time in the signal - you can hear pitch changing with each beep - that could be nice for future automated audio based checks - one can take a beep and check its running time and if it's where it's supposed to be - that's something we can't get with periodic signal.

test/file.go
```diff
@@ -49,7 +49,6 @@ func (r *Runner) testFile(t *testing.T) {
 				fileOptions: &fileOptions{
 					filename: "r_{room_name}_{time}.mp4",
 				},
-				contentCheck: r.fullContentCheck,
```
milos-lk, line 52: these content checks weren't perfect but at least giving some basic sanity guarantees in terms of expected number of beeps and flashes - the whole set of detection is built around it. What do we do now instead?

milos-lk: ah I see verifyContent now does it automatically?

test/integration.go
```diff
@@ -68,16 +69,59 @@ func (r *Runner) run(t *testing.T, test *testCase, f func(*testing.T, *testCase)
...
+		publishVideo := test.videoCodec != "" && test.videoDelay == 0
+		mp := NewRunnerPublisher(t, r, publishAudio, publishVideo)
+		test.audioTrackID = mp.AudioTrackID
+		test.videoTrackID = mp.VideoTrackID
+
+		// Schedule delayed publishes
+		if test.audioDelay != 0 && test.audioCodec != "" {
+			time.AfterFunc(test.audioDelay, func() {
+				audioPub, _ := PublishOnRoom(t, r.room, 0, true, false)
+				test.audioTrackID = audioPub.SID()
```
milos-lk, line 85: data race?

milos-lk:
The new integration.go code doesn't seem to respect test.audioCodec / test.videoCodec when choosing what to publish?
As far as I can see It always publishes participant 0's files (Opus + H264).

Not sure how PCMA/PCMU tests are working with it...
The existing `ensureRoomForTest` still switches to a PCMU/PCMA-specific room with codec constraints, but then the publish code ignores the requested codec and publishes Opus anyway.

test/file.go
milos-lk: as far as I can see not fully - they do extract flashes but don't assert they exist or match any count. They alaso extract beeps — but don't ensure they are present, no checks for beeps/flashes spacing/alignment (was 1s).
A test could produce a completely silent, black video and still pass? or I am missing something?

test/multi_participant.go
```diff
@@ -0,0 +1,387 @@
...
+				sid := pub.SID()
+				t.Cleanup(func() { _ = p.lp.UnpublishTrack(sid) })
+				results <- pubResult{idx: p.idx, isAudio: true, pub: pub}
+			}(p)
+		}
+		if p.videoTrack != nil {
+			total++
+			go func(p trackPair) {
+				pub, err := p.lp.PublishTrack(p.videoTrack, &lksdk.TrackPublicationOptions{Name: p.videoFile})
+				require.NoError(t, err)
```
milos-lk, line 242: hm - this could result in runtime.Goexit() (on error) - don't we get suck in the for range total loop below in that case? (similar comment for audio track above)

test/file.go
frostbyte73: it's all checked

milos-lk:
not sure if I am reading correctly - `AnalyzeSync` receives per-participant specs (frequencies, tile coordinates) but never uses them, it just checks that any beep is near any flash in the full composite, so one visible participant out of three would pass. Or I missed something?

frostbyte73:
> not sure if I am reading correctly - `AnalyzeSync` receives per-participant specs (frequencies, tile coordinates) but never uses them, it just checks that any beep is near any flash in the full composite, so one visible participant out of three would pass. Or I missed something?

yeah, all the existing tests use only one publisher. Was planning on using that stuff in the sdk compositing PR

frostbyte73:
> The new integration.go code doesn't seem to respect test.audioCodec / test.videoCodec when choosing what to publish? As far as I can see It always publishes participant 0's files (Opus + H264).
> 
> Not sure how PCMA/PCMU tests are working with it... The existing `ensureRoomForTest` still switches to a PCMU/PCMA-specific room with codec constraints, but then the publish code ignores the requested codec and publishes Opus anyway.

Ah, yeah forgot to get back to that. I intended to just publish 6 participants to one room and then run all the integration tests in parallel on that room, but coordinating that across multiple github runners made it too complicated. I'll update it to respect the codecs

test/file.go
milos-lk: it's probably too late for me so my brain refuses to work - just dumping thought until tomorrow when l will look at this with fresh eyes - but I think we’re mixing up structural verification with semantic content assertions. verify() still checks container/codecs/duration, and verifyContent() still runs the extractors, but most of the old per-test assertions about expected signal content were removed. For standard audio+video outputs, the new path mostly proves the analyzers ran, not that the expected beep/flash pattern was actually present and aligned

milos-lk: to be more specific I don't see where we do this or how we substitue it:
```
	avgFlashSpacing, err := averageSpacing(flashes)
	require.NoError(t, err)
	// 200ms is still pretty generous, should be tighter
	requireDurationInDelta(t, avgFlashSpacing, time.Second, time.Millisecond*200)
```

test/sync_analyzer.go
```diff
@@ -0,0 +1,786 @@
...
+		line := scanner.Text()
+
+		if m := rePTS.FindStringSubmatch(line); len(m) == 2 {
+			if d, perr := parsePTSSecondsToDuration(m[1]); perr == nil {
+				curPTS = d
+			}
+			continue
+		}
+
+		if m := reYMAX.FindStringSubmatch(line); len(m) == 2 {
```
milos-lk, line 223: hm - existing test content checks are using average, not max. API doc on top indicates legacy media (which I assume is the current test content we use). What's the reason for using max here? (a single bright enough pixel could light this up)?

milos-lk: I see - the new content doesn't have LK logo and is all dark - so YMAX is fine for new content. timer flashes are the only thing that would cross that boundary 👍

milos-lk: well it's less specific, but probably good enough

test/sync_analyzer.go
```diff
@@ -0,0 +1,786 @@
...
+	return events, scanner.Err()
+}
+
+// AnalyzeSync verifies A/V sync in a composited output file.
+//
+// For every audio beep detected from any participant, it checks that every
+// visible participant's video tile had a flash within the tolerance window.
+// Audio beeps that fall within 1 second of a mute/unmute transition are
+// skipped since propagation delay makes them unreliable.
+func AnalyzeSync(
```
milos-lk, line 632: nit: cyclomatic complexity of this function is ~18 - not super easy to follow - claude suggestion for improvment:
```
 The forward and reverse checks are essentially duplicated code. Simplification:                  
                                                                                                   
  func AnalyzeSync(...) []SyncResult {                                                             
      // ... extract beeps and flashes ...                                                         
   
      skipBefore := 5500 * time.Millisecond                                                        
      transitions := buildTransitionWindows(n, turnDuration, 1*time.Second)                      
                                                                                                   
      // Merge both directions into one check                                                      
      pairs := matchEvents(beepTimestamps, flashTimes, skipBefore, transitions)                    
                                                                                                   
      assertSyncQuality(t, pairs, tolerance)                                                       
                                                                                                   
      return results                                                                               
  }                                                                                              

  func matchEvents(setA, setB []time.Duration, ...) []eventPair { ... }                            
   
  Extract isInSkipWindow into a method on a skipWindows type, extract the "find closest in sorted  
  slice" into a helper (use sort.Search instead of linear scan), and deduplicate the             
  forward/reverse loops into a single matchEvents function called twice with swapped arguments.    
                                                                                                 
  That would drop it to roughly 5-6 per function.
```

test/samples/participant_0_440hz.ogg
milos-lk: I think we discussed with @anunaym14 using them for non egress tests as well.

PR #1198, opened 2026-04-23
new sync engine

frostbyte73:


boks1971 (approved):

biglittlebigben (approved):

pkg/pipeline/builder/audio.go
```diff
@@ -80,28 +89,14 @@ func (a *audioPacer) start(drift time.Duration) {
...
 		rate = 1 - a.tempoAdjustmentRate
 	}
-	compensationFactor := 1 / a.tempoAdjustmentRate
-	driftNanoseconds := int64(drift)
-	compensationNanoseconds := int64(compensationFactor * float64(driftNanoseconds))
-	compensationDuration := time.Duration(compensationNanoseconds)
 
-	a.remaining = compensationDuration.Abs()
-	logger.Debugw("starting audio pacer", "remaining", a.remaining, "rate", rate)
+	a.inputAtStart = a.inputAccum.Load()
```
milos-lk: data race on inputAtStart? the streaming thread from the pad probe reads it? We can make it atomic as well.

milos-lk: same for the following 2

pkg/pipeline/tempo/controller.go
```diff
@@ -7,39 +7,41 @@ import (
...
+// correction is started.
+func (tc *Controller) SetDrift(drift time.Duration) {
 	tc.mu.Lock()
-	tc.pending += drift
+	tc.drift = drift
 
 	var toStart time.Duration
-	if tc.current == 0 && tc.pending.Abs() >= DefaultThreshold {
-		// Only start if applying 'pending' keeps processed within budget.
-		if (tc.processed + tc.pending).Abs() < MaxDriftBudget {
```
milos-lk, line 39: max drift budget might still be needed - right? There might be pathological  case where we keep getting drift values in the same direction and they add up to pipeline latency - mixer will discard it if it's drifting in the past or they might get overwritten inside leaky queues if it's drifting in the future. Having some logging for that case might be useful.

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -144,44 +143,55 @@ func NewAppWriter(
...
+		timeProvider: gstreamer.NopTimeProvider(),
 	}
 	w.samplesCond = sync.NewCond(&w.samplesLock)
 
 	ts.OnKeyframeRequired = w.onKeyframeRequired
 
+	if conf.EnableSyncEngine {
+		w.trackSync.OnSenderReport(func(drift time.Duration) {
+			if w.driftHandler != nil {
+				w.driftHandler.SetDrift(drift)
```
milos-lk: Hm - I am not sure if we are sending the right value as the drift param from the SDK. 
`GetPTS` runs per packet and correctly returns wallPTS for audio when tempo controller is enabled. The onSR callback (syncengine.go:259-269) still computes `drift := sessionPTS - expectedElapsed  // ≈ 0`

This doesn't seem to use `wallClockPTS` at all. It doesn't know what PTS the audio track is actually using.  It just compares the regression output against wall clock — which are nearly equal.

So it looks like `GetPTS` is doing the right thing (wall clock for audio). But the drift signal fed to the tempo controller doesn't reflect that `GetPTS` is using wall clock? The drift signal needs to measure the gap between what `GetPTS `is producing (wall clock) and what it should be producing (NTPregression).

pkg/pipeline/tempo/controller.go
milos-lk: Maybe I missed - do we have some other mechanism to prevent this from happening?

frostbyte73: Updating in https://github.com/livekit/egress/pull/1239 - if we see it logging the hard cap we can try adding the full source reset

PR #1210, opened 2026-05-06
fix image cleanup (delete orphans)

frostbyte73:


.github/workflows/test-cleanup.yaml
```diff
@@ -76,5 +83,65 @@ jobs:
             PAGE=$((PAGE + 1))
           done
 
+          # Collect all active (tagged) digests to preserve
+          ACTIVE_DIGESTS=""
+          PAGE=1
+          while true; do
+            RESPONSE=$(curl -sf "https://hub.docker.com/v2/repositories/${REPO}/tags?page_size=100&page=${PAGE}" \
```
milos-lk, line 90: I see it's also used above - but isn't this endpoint deprecated:
https://docs.docker.com/reference/api/hub/deprecated/

PR #1212, opened 2026-05-07
Set time provider only after pipeline reaches playing state

milos-lk:
RunningTime() returns system uptime before pipeline reaches PLAYING, poisoning track synchronizer deadline if callback invoked before the pipeline transitioned to playing. When the GStreamer pipeline transitions to PLAYING, appsrc elements reach PLAYING state before the pipeline itself completes the transition and sets base_time. The pushSamples playing gate (appwriter.go:497) is signaled by the appsrc state change (watch.go:285), not the pipeline state change, so packets begin flowing through GetPTS while Pipeline.RunningTime() still computes clockTime - (unset base_time) = node monotonic uptime. getExternalMediaDeadline() in the synchronizer caches externalMediaStartTime on its first successful RunningTime() call and never corrects it. Once poisoned with a value like ~208,767 seconds (58h node uptime), the deadline is permanently ~58 hours ahead of the actual track PTS. After cMaxTimelyPacketAge (10s) of every packet being "behind" this impossible deadline, the synchronizer fires the "correcting PTS to pull the track forward" path, producing a correctedPTS equal to the node uptime rather than a sane media timestamp.

Was thinking about automation testing but it would require significant test code addition to simulate this edge case of an edge case so consciously skipping to avoid blowing the complexity out of proportion.

### DIFF
--- a/pkg/pipeline/controller.go
+++ b/pkg/pipeline/controller.go
@@ -234,9 +234,6 @@ func (c *Controller) BuildPipeline() error {
 	p.UpgradeState(gstreamer.StateStarted)
 
 	c.p = p
-	if timeAware, ok := c.src.(source.TimeAware); ok {
-		timeAware.SetTimeProvider(p)
-	}
 	close(c.callbacks.BuildReady)
 	return nil
 }
--- a/pkg/pipeline/watch.go
+++ b/pkg/pipeline/watch.go
@@ -274,6 +274,11 @@ func (c *Controller) handleMessageStateChanged(msg *gst.Message) {
 
 				logger.Infow("pipeline playing", "timeToPlaying", timeToPlaying)
 				c.updateStartTime(c.src.GetStartedAt())
+
+				// base_time is only valid after the pipeline reaches PLAYING
+				if timeAware, ok := c.src.(source.TimeAware); ok {
+					timeAware.SetTimeProvider(c.p)
+				}
 			})
 		}
 		return


pkg/pipeline/controller.go
```diff
@@ -234,9 +234,6 @@ func (c *Controller) BuildPipeline() error {
 	p.UpgradeState(gstreamer.StateStarted)
 
 	c.p = p
-	if timeAware, ok := c.src.(source.TimeAware); ok {
```
milos-lk, line 237: all consumers of these callbacks won't have it available until pipeline reaches playing state and will skip any processing based on it.

pkg/pipeline/watch.go
```diff
@@ -274,6 +274,11 @@ func (c *Controller) handleMessageStateChanged(msg *gst.Message) {
 
 				logger.Infow("pipeline playing", "timeToPlaying", timeToPlaying)
 				c.updateStartTime(c.src.GetStartedAt())
+
+				// base_time is only valid after the pipeline reaches PLAYING
+				if timeAware, ok := c.src.(source.TimeAware); ok {
+					timeAware.SetTimeProvider(c.p)
```
milos-lk, line 280: at this point we are sure pipeline reached playing state so we can feed synchronizer and track workers with it.

boks1971 (approved):

biglittlebigben (approved):
What happens in the time interval while we are playing but have no time provider?

milos-lk:
> What happens in the time interval while we are playing but have no time provider?

Definition of playing is the thing here - appsrc will transition to playing and start the whole read / push rtp packets machinery. If time provider isn't set - we simply won't be performing any adjustments - it's not mandatory input - we use it inside synchronizer to detect RTP clock skew and course correct, for that brief time interval when we don't have it set we won't be able to perform that adjustment. As soon as the entire pipeline switches to playing it's going to be armed and we could react on these.

PR #1213, opened 2026-05-08
Reintroduce sdk logs filtering

milos-lk:
Prevent sdk logs bubbling up as  errors and poluting app error logging space with SDK logs. This has been removed recently while logging refactoring but is still needed to clear the noise from egress app logging.


boks1971 (approved):

frostbyte73 (approved):
would like to figure out where these come from and intercept before this point, but that can be done later

PR #1215, opened 2026-05-08
AV sync

frostbyte73:


test/layout_test.go
```diff
@@ -0,0 +1,113 @@
...
+	// Stage x origin must be beyond the carousel (> 300px from left).
+	assert.Greater(t, stage.Rect.Min.X, 300, "stage should start well to the right")
+
+	// All thumbnails must be on the left side (< 350px wide, x < 350).
+	for i := 1; i < participants; i++ {
+		thumb := regions[i]
+		assert.Equal(t, "thumb"+string(rune('0'+i-1)), thumb.Name)
+		thumbW := thumb.Rect.Max.X - thumb.Rect.Min.X
+		assert.Less(t, thumbW, 350, "thumbnail %d width should be < 350px", i)
+		assert.Less(t, thumb.Rect.Max.X, 350, "thumbnail %d should be in carousel column", i)
```
boks1971: Can we change these to use require package for consistency?

test/content_checks.go
```diff
@@ -19,128 +20,723 @@ import (
...
+//     seconds, so frac(lag) = (1s - frac(first_stable_event_PTS)) mod 1s.
+//   - Integer part from the event count deficit: each beep that was
+//     emitted but landed before recording started is one second of
+//     integer lag. We compare total expected events (computed from
+//     per-participant windows so rotation mute/unmute doesn't inflate
+//     the count) to total detected events.
+//
+// SDK tests typically have lag < 500ms; Chrome-rendered tests
+// (RoomComposite / Template) can hit 2-3s while the page loads.
+func detectRecordingLag(eventPTS []time.Duration, plan *publishPlan, dur time.Duration) time.Duration {
```
milos-lk: Is `expectedTotal - len(sorted)` reliable as startup lag? It seems like detector misses anywhere in the file (we have seen legitimate bugs with silent 2s mid recording) would be interpreted as leading recording offset, which then shifts all expected windows. Should integer lag be based only on leading missing cadence before the first stable event?

frostbyte73: not really, but in that case the test would fail either way. Tightening the av sync tolerances makes a bunch of the tests flaky, so I'm keep it all pretty loose for now and will use the stats (next PR) to see what the problems are, then hopefully tighten things with the new sync engine

frostbyte73: improvements could be made on the publishing side as well

milos-lk: Loose tolerances make sense to me My concern is narrower: this lag estimate is not just relaxing an assertion, it changes the expected windows for the rest of the checks.

milos-lk (approved):

milos-lk (approved):

PR #1238, opened 2026-06-01
namespace -> gh

frostbyte73:
Namespace appears to have problems pulling from git lfs now


boks1971 (approved):

.github/workflows/publish-egress.yaml
boks1971, line 84: can this be read-only? Would be good to add `permissions` block.

PR #1239, opened 2026-06-01
two-tiered pacer

frostbyte73:
The old pacer/tempo controller had a hard limit on how much it could adjust for drift, to prevent something from consistently drifting in one direction until the audio mixer dropped it. In that scenario, the track would be permenently out of sync - thIs will instead ramp up the adjustment rate if the track starts drifting too far away

### DIFF
--- a/pkg/pipeline/builder/audio.go
+++ b/pkg/pipeline/builder/audio.go
@@ -312,7 +312,7 @@ func (b *AudioBin) addAudioAppSrcBinLocked(ts *config.TrackSource) error {
 				// Hard tier means the pacer cannot absorb drift fast enough even with
 				// boosting; downstream is approaching mixer alignment-threshold and
 				// will start dropping
-				logger.Errorw("audio drift exceeded hard budget",
+				logger.Warnw("audio drift exceeded hard budget",
 					errors.New("uncorrected drift above hard budget"),
 					"trackID", trackID, "tier", tier)
 				pacer.boost()
--- a/pkg/pipeline/builder/audio.go
+++ b/pkg/pipeline/builder/audio.go
@@ -292,12 +292,8 @@ func (b *AudioBin) addAudioAppSrcBinLocked(ts *config.TrackSource) error {
 	}
 
 	if pacer != nil && ts.TempoController != nil {
-		ts.TempoController.OnDriftDetectedCallback(func(drift time.Duration) {
-			if pacer.pitch != nil {
-				logger.Debugw("starting audio pacer to cover the drift", "drift", drift)
-				pacer.start(drift)
-			}
-		})
+		// tc must be set before callbacks: both replay current state synchronously.
+		pacer.tc = ts.TempoController
 		trackID := ts.TrackID
 		ts.TempoController.OnTierChange(func(tier tempo.Tier) {
 			switch tier {
@@ -318,7 +314,10 @@ func (b *AudioBin) addAudioAppSrcBinLocked(ts *config.TrackSource) error {
 				pacer.boost()
 			}
 		})
-		pacer.tc = ts.TempoController
+		ts.TempoController.OnDriftDetectedCallback(func(drift time.Duration) {
+			logger.Debugw("starting audio pacer to cover the drift", "drift", drift)
+			pacer.start(drift)
+		})
 	}
 
 	return nil
@@ -538,10 +537,6 @@ func addAudioConverter(b *gstreamer.Bin, p *config.PipelineConfig, channel livek
 }
 
 func installPitchProbes(pacer *audioPacer) {
-	if pacer.pitch == nil {
-		return
-	}
-
 	// Sink pad: accumulate input buffer durations.
 	if sinkPad := pacer.pitch.GetStaticPad("sink"); sinkPad != nil {
 		sinkPad.AddProbe(gst.PadProbeTypeBuffer, func(_ *gst.Pad, info *gst.PadProbeInfo) gst.PadProbeReturn {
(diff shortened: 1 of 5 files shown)


biglittlebigben:
Could you provide some context in the PR description?

pkg/pipeline/builder/audio.go
```diff
@@ -394,6 +298,26 @@ func (b *AudioBin) addAudioAppSrcBinLocked(ts *config.TrackSource) error {
...
+				pacer.brake()
+			case tempo.TierSoft:
+				logger.Warnw("audio drift exceeded soft budget, boosting pacer rate", nil,
+					"trackID", trackID, "tier", tier)
+				pacer.boost()
+			case tempo.TierHard:
+				// Hard tier means the pacer cannot absorb drift fast enough even with
+				// boosting; downstream is approaching mixer alignment-threshold and
+				// will start dropping
+				logger.Errorw("audio drift exceeded hard budget",
```
biglittlebigben: I think error may be too high a priority here. We typically avoid error for bad input as we currently alert on them

pkg/pipeline/builder/audio_pacer.go
```diff
@@ -0,0 +1,164 @@
...
+	outputAccum atomic.Int64 // cumulative output buffer durations (nanoseconds)
+
+	// snapshot is the active correction's baseline. nil when idle.
+	snapshot atomic.Pointer[pacerSnapshot]
+}
+
+// pacerSnapshot is the per-correction baseline captured at start(). It is
+// stored atomically as a unit so the probe never observes a torn snapshot
+// even if start() is mis-invoked while a correction is still active. The
+// active check in start() prevents this today, but treating the snapshot as
```
biglittlebigben: This is helpful insight for PR review but may not belong as is in the final commit?

biglittlebigben (commented):
A few nits. Seems sensible to me but I'll let Milos comment as well.

pkg/pipeline/builder/audio.go
```diff
@@ -394,6 +298,26 @@ func (b *AudioBin) addAudioAppSrcBinLocked(ts *config.TrackSource) error {
...
+		trackID := ts.TrackID
+		ts.TempoController.OnTierChange(func(tier tempo.Tier) {
+			switch tier {
+			case tempo.TierNormal:
+				logger.Infow("audio drift back inside soft budget", "trackID", trackID, "tier", tier)
+				pacer.brake()
+			case tempo.TierSoft:
+				logger.Warnw("audio drift exceeded soft budget, boosting pacer rate", nil,
+					"trackID", trackID, "tier", tier)
+				pacer.boost()
```
milos-lk, line 306: this somehow actives my race condition radar - need to check in more depth but probe and SR thread ops on pacer could race as far as I can see.

milos-lk:
```
  time   probe thread                            SR thread
  ────   ─────────────────────────────────       ──────────────────────────────────
  t=0                                            pacer.boost()
                                                   active.Load() → true ✓
                                                   snap := snapshot.Load()
                                                   rate := 1.20
                                                   | (preempted / gst lock contended)
  t=1    pacer.stop()                              |
           active.Swap(false) → true ✓             |
           pitch.SetArg("tempo","1.0") ✓           |
           return true                             |
         tc.DriftProcessed(actual)                 |
           ... tier drops to Normal ...            |
           tierCB → pacer.brake()                  |
             active.Load() → false ✗  no-op        |
  t=2                                              pitch.SetArg("tempo","1.20") ✗
                                                     writes "1.20"
```

Does this result in pitch speeding up audio indefinitely until next start overrides it?

pkg/pipeline/builder/audio.go
```diff
@@ -394,6 +298,26 @@ func (b *AudioBin) addAudioAppSrcBinLocked(ts *config.TrackSource) error {
...
+			case tempo.TierHard:
+				// Hard tier means the pacer cannot absorb drift fast enough even with
+				// boosting; downstream is approaching mixer alignment-threshold and
+				// will start dropping
+				logger.Warnw("audio drift exceeded hard budget",
+					errors.New("uncorrected drift above hard budget"),
+					"trackID", trackID, "tier", tier)
+				pacer.boost()
```
milos-lk, line 314: should we escalate corrective actions here - and reach for PTS snapping?

pkg/pipeline/builder/audio.go
```diff
@@ -394,6 +298,26 @@ func (b *AudioBin) addAudioAppSrcBinLocked(ts *config.TrackSource) error {
...
+			}
+		})
 		pacer.tc = ts.TempoController
```
milos-lk: `pacer.tc` is assigned after `OnDriftDetectedCallback` is registered. That callback can fire synchronously if a correction is already armed, so `start()` can run before `pacer.tc` is set and miss the current `tc.Tier()` check.

Separately, `OnTierChange` doesn't replay the current tier on registration. If the controller is already in `TierSoft`/`TierHard` at registration time, the pacer may not learn that until the next tier transition, which may not happen for a stable elevated-drift case. The next correction can then run at the base rate exactly when boosted rate is most needed.

Could we either assign `pacer.tc` before registering callbacks, or have `OnTierChange` invoke the new callback with the current tier on registration, mirroring `OnDriftDetectedCallback`?

pkg/pipeline/builder/audio.go
frostbyte73: That I'm not sure about - if there is actually a consistent drift on a track, eventually it'll get too far off from the wall clock duration, and the sync engine will force reset it. Since the PTS from the sync engine should always be within a couple seconds of the current pipeline duration, this shouldn't even be hit in the first place

pkg/pipeline/builder/audio.go
frostbyte73: should be fixed now

pkg/pipeline/builder/audio.go
milos-lk: cool - we will see that in logs anyways 👍

milos-lk (approved):

PR #1244, opened 2026-06-03
Detect wedged RTMP sinks and decouple Reset from peer ack behavior

milos-lk:
The prod symptom was an egress sitting on EGRESS_ACTIVE for ~1 hour with two RTMP outputs that had both wedged silently — Mux closed the connection, the egress burst-retried ~34 publish commands in 8 seconds, then rtmp2sink stopped emitting errors entirely. With no further error messages there was nothing to drive Reset(), so the stream stayed attached and reporting active despite no data flowing.

Adds an active liveness monitor per Stream that polls OutBytesTotal every 5s. If outbound bytes fail to grow past livenessProgressThreshold (4 KiB) for livenessIdleTimeout (30s) on a connection that has streamed real data, the monitor posts a synthetic error on the rtmp2sink element  and the existing streamFailed path tears the stream down. Each stream has its own monitor goroutine; per-stream failures are isolated so multi-output egresses keep streaming on the surviving outputs.

Also fixes a related bad-URL fast-fail bug in Reset(). The previous check relied on OutBytesAcked == 0 to detect "never connected", but OutBytesAcked only advances when the peer sends RTMP Acknowledgement messages — a peer-dependent behavior. Servers with large ack windows (or implementations that defer acks) would leave OutBytesAcked at 0 even during long healthy streams, causing the egress to incorrectly fast-fail on the first disconnect. Reset now uses a sticky monitorObservedProgress flag instead: set by the monitor whenever it observes OutBytesTotal growth above the threshold (independent of peer ack behavior), it stays true once the stream has demonstrably published real media.

Switches the monitor signal from OutBytesAcked to OutBytesTotal for the same reason — relying on peer acks for our own liveness check made the monitor dependent on the peer's ack policy. OutBytesTotal stalls under all the failure modes we care about (TCP wedge, sink wedge,  remote close) via TCP backpressure, and excludes the raw RTMP  handshake from the count.

Removes the connectedOnce / reconnectGen / 10s AfterFunc callback the monitor now provides the single source of truth for "this stream has made real progress.

Also adding toxy porxy to simulate TCP level connection disruption and tinny rtmp server to simulate stalls on protocol level. 2 new RTMP tests are added which are making use of it.

I plan adding more changes for exponential backoff when retrying but it was too much for this PR - so that should come independently.


pkg/pipeline/builder/stream.go
```diff
@@ -302,6 +353,100 @@ func (s *Stream) Stats() (*logging.StreamStats, bool) {
...
+	s.monitorOnce.Do(func() {
+		s.stopMonitor = make(chan struct{})
+		s.lastProgressAt.Store(time.Now())
+		go s.runMonitor()
+	})
+}
+
+// StopMonitor signals the monitor goroutine to exit. Idempotent.
+func (s *Stream) StopMonitor() {
+	if s.stopMonitor == nil {
```
biglittlebigben: Can we guarantee that Start and StopMonitor called from the same goroutine? If not this is a data race

pkg/pipeline/builder/stream.go
```diff
@@ -302,6 +353,100 @@ func (s *Stream) Stats() (*logging.StreamStats, bool) {
...
+func (s *Stream) checkLiveness() {
+	stats, ok := s.Stats()
+	if !ok {
+		return
+	}
+	s.applyLiveness(stats.OutBytesTotal, time.Now())
+}
+
+func (s *Stream) applyLiveness(current uint64, now time.Time) {
+	last := s.lastOutBytesTotal.Load()
```
biglittlebigben: Why does this field need to be an atomic since it seems to be only accessed by this function? If I missed another access, don't the read and the write need to be atomic, requiring an actual lock, or a "swap"?
Similar question for lastProgressAt.

pkg/pipeline/builder/stream.go
```diff
@@ -47,6 +67,26 @@ type Stream struct {
...
+
+	// liveness monitor — detects silently-wedged sinks that stop emitting both data and errors.
+	//
+	// We track OutBytesTotal (not OutBytesAcked) because the latter only advances when
+	// the peer sends RTMP Acknowledgement messages, which is peer-dependent: a server
+	// with a large ack window (e.g. some CDNs) may not ack for many seconds even when
+	// streaming is healthy. OutBytesTotal grows whenever rtmp2sink writes chunks to
+	// the socket; TCP backpressure ensures it stalls under the same failure modes we
+	// care about (TCP wedge, internal sink wedge, remote close).
+	lastOutBytesTotal       atomic.Uint64
```
biglittlebigben: Does this need to be an unsigned integer? Using a signed one would make some of the arithmetic more straightforward.

pkg/pipeline/builder/stream.go
milos-lk: Yeah - it's possible although Start is logically preceding the Stop the race can happen if we have global pipeline shutdown during these - which might call Stop() - need to protect against that - good catch

pkg/pipeline/builder/stream.go
milos-lk: 👍 leftover - doesn't need to be atomic - updating that/

pkg/pipeline/builder/stream.go
milos-lk: yeah - I used it to match the original type - but int64 will do just as fine, updating.

milos-lk: updated.

pkg/pipeline/builder/stream.go
milos-lk: updated.

pkg/pipeline/builder/stream.go
milos-lk: updated.

biglittlebigben (approved):

PR #1245, opened 2026-06-04
deps: bump protocol; drop removed media.Data ref

paulwe:
## Summary

- Bump `livekit/protocol` (jwt/v5, xsync/v4, stdlib-errors) plus transitive LiveKit pseudo-versions.
- `pkg/config/pipeline.go`: remove the `media.Data` guard — the field was deleted upstream in the protocol cleanup, so the check no longer compiles.
- `pkg/config/urls.go`: add missing `encoding/json` import surfaced by the rebuild.

Upstream: livekit/protocol#1606


pkg/config/pipeline.go
```diff
@@ -498,11 +498,6 @@ func (p *PipelineConfig) Update(request *rpc.StartEgressRequest) error {
 			p.RequestType = types.RequestTypeMedia
 			p.SourceType = types.SourceTypeSDK
 
-			// data config not yet supported
```
biglittlebigben, line 501: Why removing this?

pkg/stats/monitor.go
```diff
@@ -833,3 +834,26 @@ func (m *Monitor) checkMemoryKill(maxMemoryEgress string, maxMemoryGroup *hwstat
 		m.highMemoryStart = time.Time{}
 	}
 }
+
+// atomicFloat64 stores a float64 atomically via its IEEE-754 bit pattern.
+type atomicFloat64 struct {
```
biglittlebigben: We are redefining this in every package. It is a small piece of code that we are unlikely to have to fix across all services, but still wondering about the reason to go that route.

paulwe: i only caught this after i started reviewing the diffs. i didn't realize how bad the stdlib ergonomics are... i'm reverting this

pkg/config/pipeline.go
paulwe: this field no longer exists

biglittlebigben (approved):

PR #1251, opened 2026-06-07
Silently exit on duplicate-identity disconnect

milos-lk:
Duplicate identity for egress is an indication of faulty integrity  checks in the system - service orchestrating egress workers should ensure these don't happen at first place. If this happens - it means that a new egress recording with the existing egressID is started while the one receiving duplicate identity leave request is still active and about to switch to completed state. Switching state to completed is confusing since there is an active recording instance, still recording the stream. There is no perfect way to reconcile consistent state once we reach that point. This set of changes is aimed at minimizing inconsistencies by ignoring updates from the wroker which is about to be disconnected due to the duplicate identity reason. The new worker will keep owning the state for the affected recording from that point on.


pkg/handler/handler.go
```diff
@@ -156,11 +156,32 @@ func (h *Handler) Run() {
...
 		EgressId: egressID,
 		Metrics:  m,
 		Info:     res,
-	})
+	}
+
+	// If a duplicate participant joined with our identity and evicted us from
+	// the room, exit without writing a terminal update — another worker is
+	// running the same egress and still owns the session.
+	if h.controller.IsDuplicateIdentity() {
```
boks1971, line 168: for my education, what happens to the media this session is holding?

pkg/ipc/ipc.proto
```diff
@@ -37,6 +37,10 @@ message HandlerFinishedRequest {
   string egress_id = 1;
   string metrics = 2;
   livekit.EgressInfo info = 3;
+  // When true, the handler is exiting without a terminal update to cloud-io
+  // (e.g. the room SDK disconnected with DuplicateIdentity). The server should
+  // suppress the terminal UpdateEgress so the surviving worker owns the recording.
+  bool silent_exit = 4;
```
boks1971, line 43: Is there some chance of the new session not establishing (note SFU kicks out the old participant early in the sequence, it is possible that the new session does not full establish) and egress end not being notified? Even if the new session does not establish, there will be a participant session close on SFU side though.

test/edge.go
```diff
@@ -566,3 +586,46 @@ func (r *Runner) testEmptyStreamBin(t *testing.T, test *testCase) {
...
+	require.True(t, r.svc.IsIdle(), "egress handler should exit after duplicate-identity eviction")
+
+	// Give a small grace period for any in-flight UpdateEgress IPC to land.
+	time.Sleep(2 * time.Second)
+
+	last := r.getUpdate(t, egressID)
+	require.NotEqualf(t, livekit.EgressStatus_EGRESS_COMPLETE, last.Status,
+		"terminal COMPLETE update should be suppressed (pre-evict status was %s)", preEvictStatus)
+	require.NotEqualf(t, livekit.EgressStatus_EGRESS_FAILED, last.Status,
+		"terminal FAILED update should be suppressed (pre-evict status was %s)", preEvictStatus)
```
boks1971: Is there a need to check that `last == prevEvictStatus` to ensure that it is silent?

boks1971 (approved):

pkg/handler/handler.go
biglittlebigben: Most likely we're ending with 2 uploads with the same filename. The latter one should overwrite the first one in most cases, but not always: segmented output with timestamp as segment filename will cause some orphan segments to be left in the customer blobstore. 

In the case of file output at least, we have the option of cancelling the upload in the case where we are disconnected for duplicate identity. It would have some risk since we don't really understand why we get in this situation in the first place...

milos-lk: good question - it will upload it and get overwritten anyways - we should skip uploading media when this happens as well.

test/edge.go
milos-lk: yes - we should check this - will update

milos-lk: updated

biglittlebigben (approved):

PR #1256, opened 2026-06-10
Detect missing keyframes and re-request PLI across video codecs

milos-lk:
When an egress subscribed to a video track with no incoming keyframe (e.g. cross-region relay drops the IDR, or publisher suppresses keyframes), the GStreamer pipeline could wedge in PAUSED — avdec_h264 errored on orphaned P-frames, never produced a decodable frame, and the egress eventually aborted. The only existing PLI triggers were one-shot at stream start and on RTP loss; a single dropped PLI or keyframe left the pipeline stuck.

This wires a codec-agnostic keyframeProbe onto the parser/depay src pad for H264, VP8, and VP9 in both transcoding and passthrough paths. It drops pre-keyframe P-frames, throttled-requests a PLI while no keyframe has been seen, and marks the first accepted keyframe with DISCONT. The probe also absorbs pts_fixer's missing-PTS restoration so source-bin parsers go through a single decision point

The change introduces parsers on the transocding path which should totally be fine but is a minor pipeline topology change.


pkg/pipeline/builder/keyframe_probe.go
```diff
@@ -0,0 +1,309 @@
...
+		p.handleMissingPTS()
+		if !p.lastSrcValid.Load() {
+			return gst.PadProbeDrop
+		}
+		// Restore from the last good PTS so downstream stays monotonic. The
+		// buffer may turn out to be a keyframe that recovers the stream — in
+		// that case the keyframe branch below clears keyframePending and lets
+		// it through; otherwise the keyframePending/keyframeSeen checks at the
+		// bottom will drop it.
+		restored := gst.ClockTime(p.lastSrcPTS.Load())
```
boks1971: I don't understand this, so please bear with me. What is missing PTS?  If there is missing PTS (which I am understanding as gaps) and if you set to last PTS, will there be a/v sync issues?

pkg/pipeline/builder/keyframe_probe.go
```diff
@@ -0,0 +1,309 @@
...
+			buffer.SetFlags(buffer.GetFlags() | gst.BufferFlagDiscont)
+			if wasPending {
+				p.logger.Debugw("keyframe pending, got one")
+			}
+		}
+		p.trackKeyframe(pts)
+		return gst.PadProbeOK
+	}
+
+	if !p.keyframeSeen.Load() {
```
boks1971: is this special flag needed? can the starting state be `keyFramePending: true`?

pkg/pipeline/builder/video.go
```diff
@@ -380,6 +386,31 @@ func (b *VideoBin) addAppSrcBin(ts *config.TrackSource) error {
...
+	if err != nil {
+		return err
+	}
+	b.probesMu.Lock()
+	b.probes[name] = probe
+	b.probesMu.Unlock()
+	return nil
+}
+
+// closeProbeLocked detaches and forgets the probe for the given source-bin
```
boks1971: method name does not have `Locked`

boks1971 (approved):

pkg/pipeline/builder/keyframe_probe.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	fields := []any{}
+	if p.lastSrcValid.Load() {
+		fields = append(fields, "lastValidPTS", time.Duration(p.lastSrcPTS.Load()))
+	}
+	if avg, count, ok := p.keyframeStats(); ok {
+		fields = append(fields, "avgKeyframeInterval", avg, "keyframesTracked", count)
+	}
+	p.logger.Warnw("parser buffer missing PTS", nil, fields...)
```
biglittlebigben: Warnw is a pretty high level. Do we only expect this to happen if something went wrong in the stream?

pkg/pipeline/builder/keyframe_probe.go
```diff
@@ -0,0 +1,309 @@
...
+	if count := len(p.keyframePTS); count > 0 {
+		if delta := pts - p.keyframePTS[count-1]; delta > 0 {
+			p.totalIntervalSum += delta
+			p.totalIntervals++
+		}
+	}
+
+	p.keyframePTS = append(p.keyframePTS, pts)
+	if len(p.keyframePTS) > keyframeHistorySize {
+		p.keyframePTS = p.keyframePTS[1:]
```
biglittlebigben: Will this cause the underlying array to grow over and over (not a circular buffer)

pkg/pipeline/builder/keyframe_probe.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	now := time.Now().UnixNano()
+	last := p.lastKeyframeRequestNS.Load()
+	if last != 0 && time.Duration(now-last) < keyframeRequestInterval {
+		return
+	}
+
+	p.onRequestPLI()
+	p.lastKeyframeRequestNS.Store(now)
```
biglittlebigben: I assume this always called for the same gstreamer thread? If not there is a race setting this here outside a lock.

pkg/pipeline/builder/keyframe_probe.go
milos-lk: not needed indeed. It's residual from merging my ages old branch with recent change - will remove it, good catch.

pkg/pipeline/builder/video.go
milos-lk: stale comment - will update

pkg/pipeline/builder/keyframe_probe.go
milos-lk: this is not that frequent - very few projects had it - and it was usually one packet breaking stream - this is the logic gstreamer community suggested on overcoming their issues with parse elements which could produce buffer without PTS set. Can also be debug - will update it. This logic is existing pts_fixer logic folded into the probe with key frame detection logic so that we don't need to install 2 probes and cross CGO boundary multiple times per packet.

pkg/pipeline/builder/keyframe_probe.go
milos-lk: it will cause allocation churn  not ideal - in place shift is better here - will update it.

pkg/pipeline/builder/keyframe_probe.go
milos-lk: Sorry @boks1971 - I should have been more explicit about this. This is also moving over the existing pts_fixer logic we introduced here: https://github.com/livekit/egress/pull/1049
Reason being - that check and key frame check operate on the same probes (parsers's) so wanted to avoid installing to 2 probes and crossing CGO boundary twice for each packet.

Based on the data we had on this - it does happen from time to time (gst parse element producing a buffer without a valid PTS - which if left unset breaks processing in muxer that comes after it). So your observation is correct - this affects A/V sync but it's a price to pay for avoiding corrupted stream - not ideal - but luckily from what we observed it's always one packet that goes missing per session.

milos-lk:
we can also remove diagnostics sink pad probe - only added to capture more info - but since we discovered the root cause of the PTS issue, it's not needed. that's another artifact of these changes sitting a long time on a dev branch - I should have cleaned this as well before opening PR - sorry for that - doing this now.

pkg/pipeline/builder/keyframe_probe.go
boks1971: Thank you for the explanation Milos.

pkg/pipeline/builder/keyframe_probe.go
milos-lk: removed

pkg/pipeline/builder/video.go
milos-lk: updated

pkg/pipeline/builder/keyframe_probe.go
milos-lk: removed

pkg/pipeline/builder/keyframe_probe.go
milos-lk: updated to do in place shift - we could do circular buffer as well but that makes it a bit more complex for reading, shifting 9 elements on a key frame should be totally fine.

pkg/pipeline/builder/keyframe_probe.go
milos-lk: I got rid of sink probe - we don't need it - all these are now accessed from within single probe and are made plain types.

boks1971 (approved):
lgtm! Thank you for the clean up + simplification, I think I understood more than 50% this time 🙂

biglittlebigben (approved):

PR #1265, opened 2026-06-17
CS-1819: Drain egress on sustained high CPU instead of hard-killing

frostbyte73:
When an egress handler is killed for sustained high CPU, give it a chance to drain its current output cleanly (close MP4 moov, flush HLS playlist, finalize upload) instead of being SIGINT'd cold. Mirror the graceful path already used for max-duration. Separately, fix the highCPUDuration counter so transient CPU spikes don't accumulate forever into eventual unwarranted kills.

---

Linear: CS-1819


pkg/config/service.go
```diff
@@ -56,6 +56,8 @@ const (
 	defaultAudioTempoControllerAdjustmentRate = 0.05
 
 	defaultMaxPulseClients = 60
+
+	defaultCpuKillGraceSec = 10
```
boks1971: is 10s default enough to finish uploads?

boks1971 (approved):

pkg/config/service.go
milos-lk: not sure if we can use these upload metrics (huge cardinality) - I was having this: https://github.com/livekit/egress/blob/108f016b63ca2b9818a85df4f106e052356e6a9c/pkg/pipeline/sink/file.go#L93
It's debug level though - I remember 99th percentile being ~30s when I was looking into this last time

pkg/stats/monitor.go
```diff
@@ -49,6 +49,7 @@ type Service interface {
 	IsIdle() bool
 	IsDisabled() bool
 	IsTerminating() bool
+	StopProcess(egressID string, reason string, endReason string)
```
milos-lk: minor: these 2 reasons are confusing me slightly - do we need endReason - or could we just pass kill reason and derive end reason?

pkg/stats/monitor_prom.go
```diff
@@ -129,13 +129,16 @@ func (m *Monitor) initPrometheus() {
...
 	ResultAborted           = "aborted"
+	ResultStoppedCPU        = "stopped_cpu"
 	ResultKilledCPU         = "killed_cpu"
 	ResultKilledOOM         = "killed_oom"
 	ResultKilledShutdown    = "killed_shutdown"
 	ResultProcessError      = "process_error"
 	ResultDuplicateIdentity = "duplicate_identity"
 )
 
+const EndReasonCPUExhausted = "CPU exhausted"
```
milos-lk: minor: should this be part of protocol? all other string values we use for sendEOS are defined there.

pkg/service/process.go
```diff
@@ -223,6 +224,27 @@ func (pm *processManager) AbortProcess(egressID string, err error) {
 	logger.Infow("aborting egress completed", "egressID", egressID)
 }
 
+// StopProcess asks the handler to drain via EOS so the recording finalizes cleanly; callers must escalate to KillProcess if the handler doesn't exit.
+func (pm *processManager) StopProcess(egressID string, reason string, endReason string) {
+	logger.Infow("stopping egress", nil, "egressID", egressID, "reason", reason, "endReason", endReason)
```
milos-lk:
```suggestion
	logger.Infow("stopping egress", "egressID", egressID, "reason", reason, "endReason", endReason)
```

milos-lk (approved):

pkg/stats/monitor.go
```diff
@@ -812,6 +802,55 @@ func (m *Monitor) updateLoadRatios(cpuLoad float64) {
...
 
+// checkCPUKill stages a graceful EOS drain on sustained high CPU and escalates to a hard kill if the grace period elapses; sub-threshold ticks reset the accumulator so transient spikes don't pile up.
+func (m *Monitor) checkCPUKill(load, maxCPU float64, maxCPUEgress string) {
+	cpuKillThreshold := defaultKillThreshold
+	if cpuKillThreshold <= m.cpuCostConfig.MaxCpuUtilization {
+		cpuKillThreshold = (1 + m.cpuCostConfig.MaxCpuUtilization) / 2
+	}
+
+	if load <= cpuKillThreshold {
+		m.highCPUDuration = 0
```
biglittlebigben: This is clearing the flag very aggressively. I'm not sure how jittery CPU stats are, but if they are maybe we should wait until we don't go over the alert threshold for a few secs before clearing it

PR #1267, opened 2026-06-19
Revise CPU cost settings in README

devniel:
Hello Livekit,

The cpu_cost example is outdated, I'm adding here all the available values and comments.

Please review it.


biglittlebigben (commented):
Thanks for the patch! May I suggest you set the [default values](https://github.com/livekit/egress/blob/6009574497a528c892f9521fa302769cee48db62/pkg/config/service.go#L32) in the example as they are typically good starting point? 

Particularly, I believe 100 pulse clients is above the threshold that would cause the pulse library to fail opening new sessions.

PR #1270, opened 2026-06-22
Make stopping egress due to a room disconnect transparent

milos-lk:
If iit happens that egress gets disconnected from a room and the sdk retry logic doesn't manage to re-establish the connection - egress transparently succeeds and an API consumer has no clue that the recording got cut in the middle of the session. The goal of these chnages is to make egress fail under these circumstances but still upload what has been recorded by that time as there was nothing wrong with the media pipeline itself. To be able to integration test this I needed to add a hook for artifically triggering the disconnect flow.

### DIFF
--- a/test/edge.go
+++ b/test/edge.go
@@ -118,6 +118,25 @@ func (r *Runner) testEdgeCases(t *testing.T) {
 				custom: r.testRoomCompositeDisconnectDuration,
 			},
 
+			// Room composite where the egress participant loses its room
+			// connection with a retryable reason. The partial recording must
+			// still be finalized and uploaded, and the egress reported FAILED
+			// so cloud can retry it.
+
+			{
+				name:        "RoomCompositeRetryableDisconnect",
+				requestType: types.RequestTypeRoomComposite,
+				publishOptions: publishOptions{
+					audioCodec: types.MimeTypeOpus,
+					audioOnly:  true,
+				},
+				fileOptions: &fileOptions{
+					filename: "retryable_disconnect_{time}",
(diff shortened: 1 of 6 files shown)


test/edge.go
```diff
@@ -118,6 +118,25 @@ func (r *Runner) testEdgeCases(t *testing.T) {
 				custom: r.testRoomCompositeDisconnectDuration,
 			},
 
+			// Room composite where the egress participant loses its room
+			// connection with a retryable reason. The partial recording must
+			// still be finalized and uploaded, and the egress reported FAILED
+			// so cloud can retry it.
```
boks1971: maybe remove reference to `cloud`?

boks1971 (approved):

biglittlebigben (commented):
It is likely that most customer's logic will assume that the asset is inexistent if the Status is not Complete. Is any of the our own logic going to break if there is an output with an Error status? How will this get billed?

milos-lk:
>  Is any of the our own logic going to break if there is an output with an Error status?

based on what I checked - I don't see anything that could break. In fact, it will give more sense to future auto retry cases - where recordings resulted from these failed sessions would be preserved automatically without API customers getting FAILED status.

> It is likely that most customer's logic will assume that the asset is inexistent if the Status is not Complete

not sure about that - egress info is part of the response regardless of the end status - I wasn't able to find a doc that would imply nothing is uploaded on FAILED. In the absence of auto-retry - it gives a customer a fair chance to re-start the recording and still preserve majority of the recorded material

biglittlebigben (approved):

Issue #1272, opened 2026-06-24
room_composite hard-fails at ~18 concurrent (well below max_pulse_clients=60): image ships PulseAudio with rlimit-nofile=256; admission control guards connections, not fds


Hemendra05:
## Summary

On self-hosted egress (`v1.13.0`), `room_composite` recordings start hard-failing under concurrency at **\~18 simultaneous recordings per node** — far below the `max_pulse_clients` admission limit (default **60**). The failures are EMFILE-driven: the bundled PulseAudio daemon runs out of file descriptors because the image ships it with PulseAudio's stock `rlimit-nofile = 256`, while egress's pulse admission control counts *connections* (calibrated to the native-protocol `MAX_CONNECTIONS` ≈ 64), **not fds**. So the fd ceiling is reached first, and recordings are admitted and then fail instead of being gracefully rejected.

I believe this is the actual root cause behind [#999](<https://github.com/livekit/egress/issues/999>), livekit/egress#1160, and [#893](<https://github.com/livekit/egress/issues/893>) — which were closed as an "unfixable PulseAudio limitation" or worked around with autoscaling. It's fixable with a one-line image change (PR linked below).

## Environment

* `livekit/egress:v1.13.0`, self-hosted (docker compose, host networking), 4 egress replicas on a 32-core / 62 GiB host
* `room_composite` → mp4 file, single-speaker layout, 720p@24fps
* Stock config (no `max_pulse_clients` / cpu_cost overrides)

## Symptoms

```
pulse/pactl.go:15  failed to list pulse info  error="Connection failure: Connection terminated"
info/io.go:178     egress_failed  error="failed to change state to PLAYING" code=500
# or, on the sink-creation path:
source/web.go:79        failed to create pulse sink  error="failed to launch pulse: exit status 1"
server/server_ipc.go:43 internal error, shutting down
```

Bursty, only under load, self-recovers as concurrency drops (so it reads as random/intermittent). Critically: **no OOM** (`OOMKilled=false`), **no segfault** in `dmesg`, `/dev/shm` ~0 used, pulse modules ∝ sinks (no leak). The failure leaves no kernel trace — the signature of EMFILE.

## Root cause

PulseAudio is launched in `build/egress/entrypoint.sh` with no fd-limit override:

```bash
pulseaudio -D --verbose --exit-idle-time=-1 --disallow-exit > /dev/null 2>&1
```

[...]

Issue #1160, opened 2026-03-26
Egress recording failures under high concurrency


Durga-7780:
**Describe the bug**
We are experiencing recording failures and corrupted audio/video during bulk concurrent LiveKit Egress sessions using Room Composite and Track Composite recording.

The issue only occurs under load (~20 concurrent calls). Under low traffic, recordings work correctly.
When multiple recordings are triggered simultaneously:
- Some recordings fail to generate completely
- Some recordings are generated but audio is corrupted/distorted
- Chrome fails to launch for some egress sessions
- WebSocket timeout occurs during room page load

**Egress Version**
We are using: **livekit/egress:v1.12.0**

**Egress Request**
We are triggering Room Composite Egress using the LiveKit Python SDK as shown below:
req = egress_proto.RoomCompositeEgressRequest(
    room_name=self.room_name,
    audio_only=True,
    file=output,
)

resp = await lkapi.egress.start_room_composite_egress(req)
self._egress_id = resp.egress_id

**Additional context**
1. Deployment: Self-hosted LiveKit + Egress
2. LiveKit Server Version: 1.9.1
3. Storage: Local disk only
4. CPU: 8 cores
5. Issue occurs only under bulk concurrency (~20 simultaneous calls)
6. Likely related to Chrome/WebRTC load saturation and resource contention
7. System works normally under low concurrency

**Logs**
failed to launch chrome {"error": "page load error: websocket url timeout reached"}
internal error, shutting down {"error": "failed to launch pulse: exit status 1"}
egress_failed {"error": "failed to launch pulse: exit status 1", "code": 500}
no response from handler
process failed {"error": "egress not found"}
egress_ending {"details": "End reason: Source closed"}

We need to support bulk calling with recording **(20+ concurrent sessions) in production**. Please suggest the recommended architecture or configuration to handle this reliably.

Durga-7780:
Thanks for the suggestion 

Following up on this issue after applying the recommended configuration.

---

### Versions

* LiveKit Server: **v1.9.1**
* Egress tested:

  * `livekit/egress:v1.11.0`
  * `livekit/egress:latest`

This issue persists across both egress versions.

---

### Configuration Applied

```yaml
room_composite:
  enable_room_composite_sdk_source: true
```

Use case:

* Audio-only SIP-based voicebot
* No video involved
* 20–40 concurrent calls

---

### Current Behavior

* Egress container starts successfully
* Redis connection is established
* However, egress does not accept jobs:

```
[...]

Issue #999, opened 2025-09-11
[BUG] Egress not accepting more recordings because of "pulse clients"


nakquest:
We have a server with plenty of CPU and Memory but egress is not allowing more recordings to happen with a reason of "pulse clients"

We are running version 1.10.0 (and have tried other versions as well)

{"level":"debug","ts":1757609045.386997,"logger":"egress","caller":"stats/monitor.go:155","msg":"cpu check","nodeID":"NE_bfSR7Z2VGkjv","clusterID":"","total":72,"available":52.51891891984956,"pending":0,"used":5.08108108015044,"activeRequests":17,"activeWeb":17,"memory":9.127460479736328,"canAccept":false,"reason":"pulse clients"}

Any help to understand and solve this would be greatly appreciated.

biglittlebigben:
This is a limitation of [pulseaudio](https://www.freedesktop.org/wiki/Software/PulseAudio/), the underlying technology used to capture audio from the headless Chrome instance. Removing this limitation would require a fix from the pulseaudio team. 

If you are recording audio only RoomComposite egresses (without any video), you can try the experimental `enable_room_composite_sdk_source` config option. This will enable a path where audio mixing is done without Chrome.

Issue #893, opened 2025-04-10
[BUG]


ss14:
I have a self hosted setup of livekit server with egress deployed on a single VM (vCPU: 16 / memory: 32 GiB). The current peak load is about ~10 audio-only rooms each with 1 user, 1 agent and 1 egress. During this peak time egress crashes and restarts many times with the following error. Would really appreciate if anyone can give pointer to resolve these recurring issues.

Setup details: I following the "VM deployment" reference documentation with following image versions.
livekit-server: v1.84
egress-server: v1.9.1
egress-config.yaml:
```
room_composite_cpu_cost: 0.4
  audio_room_composite_cpu_cost: 0.4
  max_concurrent_web: 30
  web_cpu_cost: 0.4
  track_composite_cpu_cost: 0.4
  track_cpu_cost: 0.4
```

```
2025-04-10T10:41:39.897Z	INFO	egress	redis/redis.go:142	connecting to redis	{"nodeID": "NE_Kitktsf5Uj7X", "handlerID": "EGH_s3aHeShxQfKT", "clusterID": "", "egressID": "EG_B7FqoXbAfy6E", "simple": true, "addr": "localhost:6379"}
2025-04-10T10:41:39.916Z	INFO	egress	medialogutils/cmd.go:36	pactl: Failure: Module initialization failed
	{"nodeID": "NE_Kitktsf5Uj7X", "handlerID": "EGH_s3aHeShxQfKT", "clusterID": "", "egressID": "EG_B7FqoXbAfy6E"}
2025-04-10T10:41:39.916Z	INFO	egress	medialogutils/cmd.go:36	pactl: 	{"nodeID": "NE_Kitktsf5Uj7X", "handlerID": "EGH_s3aHeShxQfKT", "clusterID": "", "egressID": "EG_B7FqoXbAfy6E"}
2025-04-10T10:41:39.917Z	ERROR	egress	medialogutils/cmd.go:63	2025-04-10T10:41:39.917Z	ERROR	egress	source/web.go:79	failed to create pulse sink	{"nodeID": "NE_Kitktsf5Uj7X", "handlerID": "EGH_s3aHeShxQfKT", "clusterID": "", "egressID": "EG_B7FqoXbAfy6E", "error": "failed to launch pulse: exit status 1"}
github.com/livekit/egress/pkg/pipeline/source.NewWebSource
	/workspace/pkg/pipeline/source/web.go:79
github.com/livekit/egress/pkg/pipeline/source.New
[...]

PR #1273, opened 2026-06-24
fix(egress): raise PulseAudio fd limit to prevent EMFILE under concurrent room-composite recordings

Hemendra05:
## Problem

The egress image starts PulseAudio with its default `rlimit-nofile = 256`. Each Chrome `room_composite` recording consumes ~13.5 fds on the pulse daemon (null-sink + client connection + internal pipes), so the daemon hits **EMFILE at ~18 concurrent recordings** — far below the `max_pulse_clients` admission limit (default 60, calibrated to the native-protocol `MAX_CONNECTIONS` of 64).

Because admission control counts pulse *connections*, not *fds*, recordings get admitted and then hard-fail rather than being gracefully rejected:

```
pulse/pactl.go:15  failed to list pulse info  error="Connection failure: Connection terminated"
source/web.go:79   failed to create pulse sink  error="failed to launch pulse: exit status 1"
```

Full analysis, measurements (`fds ≈ 9 + 13.5 × recordings`; captured `open=243 limit=256/256 sinks=17` at the failure instant), and relation to #999 / #1160 / #893 are in #1272.

## Fix

Raise both the process fd limit and PulseAudio's own `rlimit-nofile` in the entrypoint before launching the daemon. `ulimit -n` alone is insufficient — PulseAudio re-applies its `daemon.conf` default (256) via `setrlimit` regardless of the inherited limit — so both lines are required. This moves the fd ceiling out to thousands so it's no longer the binding constraint, leaving the existing connection-based admission control to gate near the real 64-connection limit.

## Testing

Verified the underlying fix — raising PulseAudio's `rlimit-nofile` from the 256 default to 65536 — on a self-hosted 4-replica deployment (`v1.13.0`). On that deployment it was applied as a `/etc/pulse/daemon.conf.d` drop-in plus a matching container `ulimit`; this PR folds the equivalent into the entrypoint so it works out of the box.

- **Before:** `/proc/$(pidof pulseaudio)/limits` showed `Max open files  256  256`; room_composite recordings failed in bursts at ~18 concurrent per replica (~3–4 bursts/day).
- **After:** `limit` reads `65536`; no room_composite pulse failures across 8 days, including peak windows.

For full transparency: I have **not** run this exact entrypoint diff through a from-scratch image build — the verification above is of the equivalent `rlimit-nofile` change in production. The script parses clean (`bash -n`).

Note: on hosts whose Docker default hard `nofile` is below 65536, operators may also need `--ulimit nofile=65536:65536` on the container; default Docker (hard = 1048576) needs no change.

### DIFF
--- a/build/egress/entrypoint.sh
+++ b/build/egress/entrypoint.sh
@@ -20,6 +20,13 @@ rm -rf /home/egress/tmp/*
 
 # Start pulseaudio
 rm -rf /var/run/pulse /var/lib/pulse /home/egress/.config/pulse /home/egress/.cache/xdgr/pulse
+# PulseAudio defaults to rlimit-nofile=256, exhausted at ~18 concurrent Chrome
+# room-composite recordings (~13.5 fds each) — well below the max_pulse_clients
+# admission limit. Raise the process limit AND PulseAudio's own cap (it re-applies
+# its daemon.conf default via setrlimit, so ulimit alone is insufficient) so fd
+# exhaustion isn't the binding constraint (see #1272).
+ulimit -n 65536 || true
+echo "rlimit-nofile = 65536" >> /etc/pulse/daemon.conf
 pulseaudio -D --verbose --exit-idle-time=-1 --disallow-exit > /dev/null 2>&1
 
 # Run egress service


biglittlebigben:
Thanks for the report. A slightly different fix was committed in https://github.com/livekit/egress/commit/b6985aec4391171f46506ec7a0a3e1bcbaf45f83

PR #1275, opened 2026-06-24
Reap orphaned PulseAudio sinks left by crashed handlers

### LINKED ISSUE #1274: PulseAudio module/client leak on crashed handlers permanently disables web egress ("pulse clients" rejection)
## Summary

On long-lived egress pods, abnormally-terminated web/room-composite handlers leak their PulseAudio `module-null-sink` (and the associated pulse clients) into the shared, pod-lifetime PulseAudio daemon. Nothing reconciles or reaps these orphans, so the live pulse-client count only ever grows. Once it crosses `MaxPulseClients`, `Monitor.canAcceptWebLocked()` returns `false` permanently and the instance rejects **all** further web/room-composite requests with `reason: "pulse clients"` — i.e. the pod becomes a "zombie" that is up and healthy-looking but can no longer accept any web-based egress.

## Version

- egress `v1.13.0` (also reproducible on `v1.12.0`)
- Deployment: one shared PulseAudio daemon per pod (`pulseaudio -D --exit-idle-time=-1 --disallow-exit`), long-lived pods serving many recordings.

## Root cause

**The only code path that unloads a sink runs inside the handler subprocess.** `WebSource.Close()` is the sole place `pactl unload-module` is ever called:

```go
// pkg/pipeline/source/web.go  (Close)
if s.pulseSink != "" {
    logger.Debugw("unloading pulse module")
    if err := exec.Command("pactl", "unload-module", s.pulseSink).Run(); err != nil {
        logger.Errorw("failed to unload pulse sink", err)
    }
}
```

The matching create runs per recording (`pkg/pipeline/source/web.go`, `createPulseSink`):

```go
pactl load-module module-null-sink sink_name="<EgressId>" sink_properties=...
```

Because the PulseAudio daemon is shared and pod-lifetime, but the unload lives in the per-recording handler behind a `sync.Once`, any handler that dies **without** running its normal shutdown — Chrome/GStreamer crash, OOM-kill, or any SIGKILL of the child — never executes `unload-module`. The orphaned `module-null-sink` and its connected clients then remain in the shared daemon **for the life of the pod**.

**Nothing reaps the orphans.** There is no reconciliation between live pulse state and active egresses. The helper that could do it — `PulseInfo.GetEgressInfo()` in `pkg/pipeline/source/pulse/pactl.go`, which maps each sink back to its EgressId — is never used for cleanup.

**The leak then saturates admission.** `pulse.Clients()` counts every client on the daemon (orphans included):

```go
// pkg/pipeline/source/pulse/pactl.go
func Clients() (int, error) {
    info, err := List()
    if err != nil { return 0, err }
    return len(info.Clients), nil
}
```

and admission is gated on it:

```go
// pkg/stats/monitor.go  (canAcceptWebLocked)
func (m *Monitor) canAcceptWebLocked() bool {
    clients, err := pulse.Clients()
    if err != nil { return false }
    return clients + int(m.pendingPulseClients.Load()) + pulseClientHold <= m.cpuCostConfig.MaxPulseClients
}
```

With the defaults `MaxPulseClients = 60` (`pkg/config/service.go`) and `pulseClientHold = 4` (`pkg/stats/monitor.go`), once leaked clients push the live count past the limit, every web/room-composite admission check fails and logs:

```go
// pkg/stats/monitor.go
fields = append(fields, "canAccept", false, "reason", "pulse clients")
```

The count never recovers (daemon never cleans, only unloader is dead), so the rejection is **permanent** for that pod.

## Steps to reproduce

1. Run an egress instance that serves repeated web/room-composite (e.g. audio-only room composite) egresses on a long-lived pod.
2. Cause some handlers to terminate abnormally (SIGKILL the handler subprocess mid-recording, or drive the pod into OOM so handlers are killed before `Close()` runs).
3. Observe in `pactl list short modules` / `pactl --format json list` that `module-null-sink` entries named after old EgressIds accumulate and are never removed, and that `pactl list clients` grows over time.
4. After enough orphans accumulate (live client count approaches `MaxPulseClients`), every new web/room-composite request is rejected with `canAccept=false reason="pulse clients"`, while the process otherwise stays healthy.

## Impact

- **Silent, permanent capacity loss.** Affected pods stop accepting web/room-composite egress entirely but report no error to standard health checks (the `/healthz`-style liveness on the metrics port returns 200 regardless of pulse state), so the orchestrator never restarts them.
- A single, *expected* handler crash can escalate into a pod that is permanently unusable for web egress, requiring external detection (log-scraping for `"pulse clients"`) and manual/automated pod deletion to recover.

## Suggested fix

Reconcile pulse state against active egresses and reap orphans, rather than relying solely on the handler's `Close()`:

- Add a periodic reaper (the monitor already ticks via `updateEgressStats` and already calls `pulse.Clients()` in `updateLoadRatios`). On each tick, list sinks via `pulse.List()`, and for any `module-null-sink` whose `sink_name` (EgressId) is not in the set of active/pending egresses for more than a small grace period, `pactl unload-module` it. `GetEgressInfo()` already provides the sink→EgressId mapping.
- A grace window avoids reaping a just-created sink before its handler registers.

This makes a crashed handler self-heal instead of permanently poisoning the pod, and removes the need for external "pulse clients" log-scraping watchdogs.

Happy to open a PR implementing the reaper if the approach looks reasonable.

biswajitpain:
## What

Reap orphaned PulseAudio null-sinks left behind by egress handlers that died without running their normal cleanup, so a crashed recording self-heals instead of permanently disabling web egress on the pod.

Fixes #1274.

## Why

Each web/room-composite handler loads a `module-null-sink` (named after its EgressId) into the shared, pod-lifetime PulseAudio daemon, and only unloads it in `WebSource.Close()`. When a handler dies abnormally (Chrome/GStreamer crash, OOM-kill, SIGKILL), `Close()` never runs, so the sink and its clients leak on the daemon. Nothing reconciles them. Over time the live pulse-client count climbs until `Monitor.canAcceptWebLocked()` permanently returns false and the instance rejects every web/room-composite request with `reason: "pulse clients"` — a zombie pod that standard health checks don't catch.

See #1274 for the full root-cause write-up.

## How

- Adds a periodic reaper to the monitor, hooked into the existing `updateEgressStats` tick (which already calls into pulse).
- Each tick it lists pulse sinks and unloads any egress-created null-sink (name prefixed with `utils.EgressPrefix`) whose EgressId is **not** in the active set (`pending` ∪ running `procStats`) and has been orphaned longer than a grace period.
- A grace window (`pulse_sink_reap_grace_sec`, default 30s) plus first-seen tracking avoids reaping a sink during the normal teardown gap between `EgressEnded` and the handler's own unload, and never touches the base `auto_null` sink or non-egress sinks.
- `pactl unload-module` calls are dispatched off the monitor lock.
- Gated by `reap_orphaned_pulse_sinks` (default enabled); set to `false` to restore prior behavior.

## Config

```yaml
cpu_cost:
  reap_orphaned_pulse_sinks: true   # default
  pulse_sink_reap_grace_sec: 30     # default
```

## Testing

- New unit tests (`pkg/stats/monitor_pulse_reap_test.go`) cover the decision logic without requiring `pactl`: active/base/foreign sinks are ignored; orphans are reaped only after the grace period; tracking clears when a sink becomes active again or disappears (handler cleaned it up itself).
- `go test ./pkg/stats/...`, `go build`, `go vet`, and `gofmt` all pass.

I'm happy to adjust the approach (e.g. default-off, different grace handling, or moving the reaper to its own ticker) based on review.

### DIFF
--- a/pkg/config/service.go
+++ b/pkg/config/service.go
@@ -58,6 +58,8 @@ const (
 	defaultMaxPulseClients = 60
 
 	defaultCpuKillGraceSec = 30
+
+	defaultPulseSinkReapGraceSec = 30
 )
 
 type ServiceConfig struct {
@@ -93,6 +95,13 @@ type CPUCostConfig struct {
 	TrackCompositeCpuCost     float64 `yaml:"track_composite_cpu_cost"`
 	TrackCpuCost              float64 `yaml:"track_cpu_cost"`
 	MaxPulseClients           int     `yaml:"max_pulse_clients"` // pulse client limit for launching chrome
+
+	// Reaping of orphaned PulseAudio null-sinks left behind by handlers that died without
+	// running their normal cleanup (crash/OOM/SIGKILL). Without this, leaked sinks/clients
+	// accumulate on the shared daemon until canAcceptWebLocked() permanently rejects with
+	// reason "pulse clients". Enabled by default.
+	ReapOrphanedPulseSinks *bool `yaml:"reap_orphaned_pulse_sinks"`
+	PulseSinkReapGraceSec  int   `yaml:"pulse_sink_reap_grace_sec"` // seconds a sink must be orphaned before it is unloaded (0 = use default)
 
 	// Memory source configuration (cgroup-aware memory accounting)
 	MemorySource       MemorySource `yaml:"memory_source"`         // memory measurement source: proc_rss, cgroup
@@ -183,6 +192,13 @@ func (c *ServiceConfig) InitDefaults() {
 	}
 	if c.CpuKillGraceSec <= 0 {
 		c.CpuKillGraceSec = defaultCpuKillGraceSec
+	}
+	if c.ReapOrphanedPulseSinks == nil {
+		enabled := true
+		c.ReapOrphanedPulseSinks = &enabled
+	}
+	if c.PulseSinkReapGraceSec <= 0 {
+		c.PulseSinkReapGraceSec = defaultPulseSinkReapGraceSec
 	}
 
 	// Memory source defaults to proc_rss (preserves existing behavior)
--- a/pkg/pipeline/source/pulse/pactl.go
+++ b/pkg/pipeline/source/pulse/pactl.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"os/exec"
+	"strconv"
 
 	"github.com/livekit/egress/pkg/errors"
 )
@@ -14,6 +15,19 @@ func Clients() (int, error) {
 		return 0, err
 	}
 	return len(info.Clients), nil
+}
+
+// UnloadModule unloads a PulseAudio module by index (e.g. the owner module of an
+// orphaned null-sink). Unloading an already-removed module returns an error, which
+// callers may treat as benign.
+func UnloadModule(index int) error {
+	var e bytes.Buffer
+	cmd := exec.Command("pactl", "unload-module", strconv.Itoa(index))
+	cmd.Stderr = &e
+	if err := cmd.Run(); err != nil {
+		return errors.New(e.String())
+	}
+	return nil
 }
 
 func List() (*PulseInfo, error) {
--- a/pkg/stats/monitor.go
+++ b/pkg/stats/monitor.go
@@ -17,6 +17,7 @@ package stats
 import (
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/linkdata/deadlock"
@@ -27,6 +28,7 @@ import (
 	"github.com/livekit/protocol/egress"
 	"github.com/livekit/protocol/logger"
 	"github.com/livekit/protocol/rpc"
+	"github.com/livekit/protocol/utils"
 	"github.com/livekit/protocol/utils/hwstats"
 
 	"github.com/livekit/egress/pkg/config"
@@ -90,6 +92,7 @@ type Monitor struct {
 	cgroupOK           bool
 	cgroupErrorLogged  atomic.Bool
 	pulseErrorLogged   atomic.Bool
+	orphanedSinks      map[string]time.Time // egress null-sink name -> first time seen without an active egress
 }
 
 type processStats struct {
@@ -115,6 +118,7 @@ func NewMonitor(conf *config.ServiceConfig, svc Service) (*Monitor, error) {
 		svc:            svc,
 		pending:        make(map[string]*processStats),
 		procStats:      make(map[int]*processStats),
+		orphanedSinks:  make(map[string]time.Time),
 		lastMemoryDump: time.Now(),
 	}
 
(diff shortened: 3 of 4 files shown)


pkg/config/service.go
```diff
@@ -94,6 +96,13 @@ type CPUCostConfig struct {
 	TrackCpuCost              float64 `yaml:"track_cpu_cost"`
 	MaxPulseClients           int     `yaml:"max_pulse_clients"` // pulse client limit for launching chrome
 
+	// Reaping of orphaned PulseAudio null-sinks left behind by handlers that died without
+	// running their normal cleanup (crash/OOM/SIGKILL). Without this, leaked sinks/clients
+	// accumulate on the shared daemon until canAcceptWebLocked() permanently rejects with
+	// reason "pulse clients". Enabled by default.
+	ReapOrphanedPulseSinks *bool `yaml:"reap_orphaned_pulse_sinks"`
```
biglittlebigben, line 103: I think this would be better in the main `ServiceConfig` structure as it is not directly a quota enforcement related configuration. We'd probably remove the conf entries entirely once made sure that this doesn't cause any regression.

pkg/config/service.go
```diff
@@ -94,6 +96,13 @@ type CPUCostConfig struct {
...
+	PulseSinkReapGraceSec  int   `yaml:"pulse_sink_reap_grace_sec"` // seconds a sink must be orphaned before it is unloaded (0 = use default)
```
biglittlebigben, line 104: Instead of a separate boo for enabling/disabling, can we use `PulseSinkReapGraceSec < 0` as the way to disable the reaper?

pkg/pipeline/source/pulse/pactl.go
```diff
@@ -16,6 +17,19 @@ func Clients() (int, error) {
 	return len(info.Clients), nil
 }
 
+// UnloadModule unloads a PulseAudio module by index (e.g. the owner module of an
```
biglittlebigben, line 20: The function is fairly self explanatory. This comment doesn't really add any useful information and would best be removed.

pkg/stats/monitor.go
```diff
@@ -706,9 +710,107 @@ func (m *Monitor) updateEgressStats(stats *hwstats.ProcStats) {
...
+	if err != nil {
+		return
+	}
+
+	for _, reap := range m.planSinkReapsLocked(info.Sinks, time.Now()) {
+		module := reap.module
+		name := reap.name
+		go func() {
+			if err := pulse.UnloadModule(module); err != nil {
+				logger.Warnw("failed to reap orphaned pulse sink", err, "sink", name, "module", module)
```
biglittlebigben, line 742: Please adjust this down to debug. Since there is a race between pulse.List() and the actual call to Unload, failures to Unload pulse can be expected.

pkg/stats/monitor.go
```diff
@@ -706,9 +710,107 @@ func (m *Monitor) updateEgressStats(stats *hwstats.ProcStats) {
 
 	m.updateLoadRatios(load)
 
+	m.reapOrphanedPulseSinksLocked()
+
 	m.checkMemoryKill(maxMemoryEgress, maxMemoryGroup)
 }
 
+// reapOrphanedPulseSinksLocked unloads PulseAudio null-sinks whose owning egress is no
```
biglittlebigben, line 718: Please keeps this comment more succinct (First sentence is probably enough)

pkg/stats/monitor.go
```diff
@@ -706,9 +710,107 @@ func (m *Monitor) updateEgressStats(stats *hwstats.ProcStats) {
...
+		}()
+	}
+}
+
+type sinkReap struct {
+	name   string
+	module int
+}
+
+// planSinkReapsLocked decides which orphaned egress null-sinks are past the grace period and
```
biglittlebigben, line 755: The goal of the function is fairly self explanatory and this comment is not needed.

biglittlebigben (commented):
Thanks for tracking down the issue and proposing a fix. Some comments (mostly nits).

biglittlebigben:
I addressed the feedback in https://github.com/livekit/egress/pull/1279. Thanks again for the patch. Could you accept the CLA so that we cab merge it.

Thanks!

PR #1290, opened 2026-07-10
Restore livekitwebrtcsrc/sink in gstreamer images, support revision tags

milos-lk:
Enable the LiveKit signaller when building gst-plugins-rs: newer upstream releases moved it behind an opt-in meson option (`webrtc-livekit`, disabled by default), so the 1.26.7 images silently lost `livekitwebrtcsrc` /  `livekitwebrtcsink` (present in 1.24.12). `compile-rs` now passes `-D webrtc-livekit=enabled` and fails the build either element is missing, so this can't regress silently again.

Decouple the docker tag from the upstream source version in the publish workflows via a new optional `image_tag` input (defaults to `version`). This allows republishing a fixed build of an already-released version under a  revision tag (e.g. `1.26.7-1`) instead of mutating the existing `1.26.7` tags. Only the base stage uses the version for source downloads; the dev/prod/prod-rs stages use the build-arg solely in `FROM` lines, so they chain off the revision-tagged base.

### DIFF
--- a/.github/workflows/publish-gstreamer-base.yaml
+++ b/.github/workflows/publish-gstreamer-base.yaml
@@ -4,6 +4,12 @@ on:
       version:
         required: true
         type: string
+      image_tag:
+        # docker tag to publish under; may differ from version for revised
+        # builds of the same upstream release (e.g. 1.26.7 -> 1.26.7-1)
+        required: false
+        type: string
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
nice!

PR #1295, opened 2026-07-13
Feature/egress h264 passthrough

curruwilla:


biglittlebigben (commented):
Thanks for the PR. We tried to implement passthrough video export in the past but ended up going away from it because without transcoding it is simply impossible to ensure that the stream abides to the required parameter:
- No way to split the stream in segments at the required boundary (other than best effort). The HLS spec particularly forbids any segment longer than the target duration https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.3.1. 
- Dealing with track muting properly is not possible. Your patch repeats the last I frame, which would yield a very awkward stream in most cases.

curruwilla:
> Thanks for the PR. We tried to implement passthrough video export in the past but ended up going away from it because without transcoding it is simply impossible to ensure that the stream abides to the required parameter:
> 
> * No way to split the stream in segments at the required boundary (other than best effort). The HLS spec particularly forbids any segment longer than the target duration https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.3.1.
> * Dealing with track muting properly is not possible. Your patch repeats the last I frame, which would yield a very awkward stream in most cases.

Thanks for taking the time to review this — you're right that a general-purpose passthrough can't guarantee segment boundaries, and that's exactly why this is behind an opt-in config flag rather than a default. But I'd like to push back on two specifics.

On the TARGETDURATION requirement: the RFC text you quoted constrains each segment's EXTINF duration, rounded to the nearest integer, against the declared target duration — not against the requested segment_duration. Nothing in the spec requires declaring the nominal segment duration as the target. This PR declares 2× the nominal duration in pass-through mode (segments.go), keeps the measured durations in #EXTINF (already the existing behavior), and logs a warning if a segment ever exceeds the declared target. Deriving TARGETDURATION from actual max segment duration rather than the requested one is what ffmpeg's HLS muxer and Shaka Packager do as well. Note the current transcode path also produces variable-length segments — splitmuxsink only cuts at keyframes even with a local encoder; the difference is overshoot magnitude (~1 frame vs one GOP + PLI RTT), not kind.

Where your concern is fully valid: if the publisher never delivers a keyframe within the headroom (ignores PLI, sustained loss), the segment exceeds the declared target and nothing can enforce it — that's a real, unbounded tail for arbitrary publishers. That's the case I'd scope this feature away from: it's intended for deployments that control the publisher (fixed GOP ≤ segment_duration), which is where the CPU win matters most. Happy to make that constraint louder in docs, or fail back to transcode at runtime when the observed GOP exceeds the headroom.

On the freeze fallback: the alternative in the current decoded path is videotestsrc pattern=black — a black screen. Repeating the last IDR gives a frozen frame, which is what most conferencing UIs show on mute, and it's a valid stream that keeps segment cuts exact (every injected frame is a keyframe). The genuinely rough edge is the unmute transition (delta frames dropped until the PLI keyframe arrives, so freeze→jump). If the freeze behavior itself is the sticking point, I'm happy to make the fallback configurable.

Would you be open to this landing as an explicitly opt-in mode with documented publisher requirements, or is the position that passthrough won't be accepted upstream in any form? If the latter, no hard feelings — I'll maintain it as a fork and would just appreciate a pointer to the earlier attempt you mentioned, to learn from what broke.

PR #1298, opened 2026-07-15
Pass handler config/request via env vars, not cmdline

biglittlebigben:
## Summary

The per-egress config and request are passed to the hidden `run-handler` child process as `--config` / `--request` command-line args. The marshaled `PipelineConfig` carries secrets — Redis password, S3/Azure/GCP credentials, API keys — so they leak to any local user via `ps` / `/proc/<pid>/cmdline`.

This moves both payloads onto the child process **environment** instead:

- `pkg/server/server_rpc.go` — `launchProcess` sets `EGRESS_HANDLER_CONFIG_BODY` and `EGRESS_HANDLER_REQUEST` on `cmd.Env` (inheriting `os.Environ()`) instead of appending CLI args.
- `cmd/server/main.go` — the `run-handler` `--config` / `--request` flags gain those env vars as `Sources`, so they're populated from the environment. `runHandler` is unchanged (`c.String(...)` resolves the env source transparently), and the flags still work for manual invocation.

Env vars also sidestep `ARG_MAX` limits on the (potentially large) config+request payloads.

Closes [CS-1878](https://linear.app/livekit/issue/CS-1878/bug-api-secret-leaks-in-egress-cmdline)

## Test plan

- [ ] `pkg/server` builds cleanly
- [ ] Existing integration suite (`mage integration`) exercises the real parent→child launch path

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -57,10 +57,12 @@ func main() {
 				Description: "runs a request in a new process",
 				Flags: []cli.Flag{
 					&cli.StringFlag{
-						Name: "request",
+						Name:    "request",
+						Sources: cli.EnvVars("EGRESS_HANDLER_REQUEST"),
 					},
 					&cli.StringFlag{
-						Name: "config",
+						Name:    "config",
+						Sources: cli.EnvVars("EGRESS_HANDLER_CONFIG_BODY"),
 					},
 				},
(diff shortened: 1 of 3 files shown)


milos-lk (approved):

biglittlebigben:
Fixing the integration test

PR #1305, opened 2026-07-19
Remove appsrc recovery on flushing

milos-lk:
The reason we introduced recovery on persistent flow flushing errors while pushing packets for a track was appsrc src pad task deactivation due to race condition which is fixed in https://github.com/livekit/egress/pull/1293

There is another reason we might get flow flushing - and that's not fully correct and forceful rmtp sink removal / re-adding on e.g rtmp connectivity issue. That issue will be fixed independently - but is now causing appsource recovery path when it happens, which doesn't really help (as it's not the original problem the recovery is added for) and might even cause deadlocks.

When persistent flow flushing is happening due to the rtmp issue - it's the most honest to fail the egress as soon as we exceeded tolerable threshold instead of silently continuing without a stream. As mentioned the rmtp source of flushing error will also be addressed independently.


boks1971 (approved):
Image  always a good thing :-)

PR #1307, opened 2026-07-22
Dedup a retried start from sdk retry

cnderrauber:


anunaym14 (approved):

pkg/server/server_rpc.go
```diff
@@ -58,6 +58,12 @@ func (s *Server) StartEgress(ctx context.Context, req *rpc.StartEgressRequest) (
 		s.activeRequests.Dec()
 		return nil, errors.ErrEgressAlreadyExists
 	}
+	// Dedup a retried start from sdk retry
+	if info, err := s.ioClient.GetEgress(ctx, &rpc.GetEgressRequest{EgressId: req.EgressId}); err == nil && info != nil {
```
milos-lk, line 62: could you comment on why we need this? This check could also be racy - it's still not atomic and suffers the from same problem as the one we discussed [here](https://github.com/livekit/livekit/pull/4694/changes#r3621698187)
and looks redundant to me. 
Since the price that we need to pay is one more RPC/db roundtrip on each start egress request - I am not sure it's worth it tbh.

milos-lk:
This will also break server side cloud-egress auto retry.

milos-lk (changes requested):

pkg/server/server_rpc.go
cnderrauber: It does have a small race window. Consider that the chance of two on-flight duplicate requests is low: it only happens when the first api call is processing on the first node that has not created record yet, then the client timeout on the first call and retries the second call arrives at the egress node. I think the small race window is acceptable here.

PR #1308, opened 2026-07-22
test: serve web egress fixture from localhost instead of blender.org

erikhortsch:
## Why

The web integration tests (`TestEgress/*/Web`) navigate headless Chrome to
`https://download.blender.org/peach/bigbuckbunny_movies/BigBuckBunny_320x180.mp4`.
Blender removed the loose `.mp4` files (only the `.zip` remains), so that URL now
**404s** — first intermittently as CDN edges expire, then permanently. The
navigation gets an HTTP ≥400, which the web source treats as a non-retryable
page-load error, so the Web subtests fail on essentially every egress CI run.

## What

Stop depending on the external URL. The clip is now vendored in
`media-samples` (see livekit/media-samples#14, git-lfs, CC BY 3.0). This change
serves it from a localhost file server rooted at the existing `/media-samples`
dir and points the web tests at that URL instead:

```go
var webURL = sync.OnceValue(func() string {
    ln, _ := net.Listen("tcp", "127.0.0.1:0")
    go http.Serve(ln, http.FileServer(http.Dir("/media-samples")))
    return fmt.Sprintf("http://%s/BigBuckBunny_320x180.mp4", ln.Addr())
})
```

Same clip, same behavior — just served locally, so the tests no longer make an
external request (and all matrix jobs stop hammering blender.org concurrently).

## Ordering / dependency

Requires the clip to be present at `/media-samples/BigBuckBunny_320x180.mp4` in
the test image — i.e. **merge livekit/media-samples#14 first**, then this. Until
then the Web tests would 404 locally instead of remotely.

## Note

Integration tests need gstreamer + the `integration` build tag, so this wasn't
compiled locally; the change is confined to `test/builder.go` (stdlib-only
additions) and CI will exercise it.

### DIFF
--- a/test/builder.go
+++ b/test/builder.go
@@ -17,7 +17,11 @@
 package test
 
 import (
+	"fmt"
+	"net"
+	"net/http"
 	"path"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


biglittlebigben (approved):

milos-lk:
Thanks @erikhortsch 💯

PR #1309, opened 2026-07-23
Serve handler RPC topics from the service process over IPC

milos-lk:
## Why

Each egress handler subprocess currently opens its own message-bus connection during startup, solely to serve three per-egress RPC topics (`UpdateStream`, `UpdateEgress`, `StopEgress`). This sits in the pre-ready window, so bus connect/registration failures or slowness surface as `handler failed to start` after the 10s launch timeout — historically the dominant transient cause of launch failures — and every concurrent handler multiplies bus connections and reconnect storms. Ingress already avoids all of this by keeping its handlers bus-free and forwarding RPCs from the main process over IPC.

## What

- **`ipc.proto`**: adds `UpdateStream` / `UpdateEgress` / `StopEgress` to the `EgressHandler` IPC service, using the same `livekit.*` request/response types as the psrpc definitions. The existing handler methods already match the generated gRPC signatures, so they become the IPC implementations unchanged. psrpc error codes survive the hop (`psrpc.Error` implements `GRPCStatus()`, and psrpc's response encoding maps status errors back symmetrically).
- **New `service.HandlerRPCProxy`**: implements `rpc.EgressHandlerServerImpl` on a single node-wide psrpc server, resolving the target handler via the process manager's IPC client per request.
- **Topic lifecycle owned by the ProcessManager**: registration happens inside `Launch` — after the IPC client exists and the process is in the active map, before `cmd.Start` — so there is no window where an advertised topic can't resolve a client. Deregistration happens in `ProcessFinished`, which runs on every exit path; `DeregisterHandler` is a safe no-op for unknown topics, making cleanup idempotent.
- **`grpc.WaitForReady(true)`** on the proxied calls: a request arriving while the handler's IPC listener is still coming up waits, bounded by the caller's psrpc request deadline, instead of failing fast. The handler-side implementations already gate on the `initialized` fuse, so a stop landing mid-boot resolves into a graceful early stop.
- **Handler slimming**: `NewHandler` drops the bus parameter and its psrpc server; `run-handler` drops the redis connection.

## Behavior notes

- Registration failure inside `Launch` is handled like a `cmd.Start` failure: error returned after `EgressStarted`, cleanup via `ProcessFinished`, accounting paired.
- Register/deregister are local buffered writes on the shared connection (no dial, no server round-trip), verified down through psrpc and the nats client; no lock is held across I/O, so a slow call affects only its own StartEgress.
- Boot-window requests change from "no subscriber → full-timeout `ErrNoResponse`" to either waiting for the handler or failing at the same deadline with the same `DeadlineExceeded` class — never a false `NotFound`.

## Testing

Full file-integration suite run on this branch: 27/31 passing, with every passing test exercising the new path end to end (bus-free handler launch, `HandlerReady` over IPC, `StopEgress` via bus → proxy → IPC). The 4 failures were `TestEgress/*/Web*`, caused by the since-removed blender.org fixture URL — fixed independently by #1308, which is now merged into this branch.

### DIFF
--- a/pkg/service/process.go
+++ b/pkg/service/process.go
@@ -53,9 +53,8 @@ type ProcessManager interface {
 	// SetHandlerTopicHooks installs callbacks used to advertise and remove the
 	// per-egress handler RPC topics. Registration happens inside Launch, after
 	// the handler's IPC client exists; deregistration happens in ProcessFinished.
-	// The hooks are stored without synchronization and read by Launch and
-	// ProcessFinished, so this must be called once during server construction,
-	// before any handler can be launched.
+	// The hooks are read without synchronization, so this must be called once
+	// during server construction, before any handler is launched.
 	SetHandlerTopicHooks(register func(egressID string) error, deregister func(egressID string))
 	GetContext(egressID string) context.Context
 	AlreadyExists(egressID string) bool
--- a/pkg/service/process.go
+++ b/pkg/service/process.go
@@ -50,6 +50,10 @@ const (
 
 type ProcessManager interface {
 	Launch(ctx context.Context, handlerID string, req *rpc.StartEgressRequest, info *livekit.EgressInfo, cmd *exec.Cmd) error
+	// SetHandlerTopicHooks installs callbacks used to advertise and remove the
+	// per-egress handler RPC topics. Registration happens inside Launch, after
+	// the handler's IPC client exists; deregistration happens in ProcessFinished.
+	SetHandlerTopicHooks(register func(egressID string) error, deregister func(egressID string))
 	GetContext(egressID string) context.Context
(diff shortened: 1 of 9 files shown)


boks1971 (approved):

pkg/service/process.go
```diff
@@ -82,6 +89,11 @@ func NewProcessManager() ProcessManager {
 	}
 }
 
+func (pm *processManager) SetHandlerTopicHooks(register func(egressID string) error, deregister func(egressID string)) {
```
biglittlebigben, line 94: Since there is no lock, this needs to be called early in the Server lifecycle before we start any handler. Is this worth a comment?

biglittlebigben (approved):
Thanks. This was on my TODO list...

PR #1313, opened 2026-07-27
Don't declare the pipeline frozen when EOS arrived before it was requested

milos-lk:
Logs are showing some VP8/H264 track recordings failing with "pipeline frozen", but the pipeline wasn't frozen, it had already finished. When the publisher left the room, their encoder emitted a degenerate final keyframe (8x8), the muxer refused the caps renegotiation, and the appsrc died with not-negotiated, pushing EOS downstream as its last act. That EOS reached the file sink and posted on the bus before the controller's own end path ran. sendEOS() then installed its sink EOS probes after the EOS event had already passed (so they could never fire) and armed the 30-second watchdog after the bus handler had already tried to stop a timer that didn't exist yet (so nothing could ever stop it). The timer fired, declared the pipeline frozen, and a correctly finalized recording was reported as failed.

Logs show this class of shutdown is common and always benign: end-of-track placeholder frames (8x8 VP8, 2x2 H264) trigger the same not-negotiated error-EOS a few times a day, and every sampled instance coincides with the publisher leaving the room, the frozen verdict is just the unlucky ordering of a race the shutdown usually wins.

The fix makes sendEOS() respect the eosReceived fuse: if the bus EOS was already processed, the pipeline is already stopping and there is nothing to arm. The watchdog callback also checks the fuse, closing the narrow window where the EOS lands between the guard and the timer assignment; in that case it falls through to the idempotent Stop() instead of failing the egress. A genuinely frozen pipeline, EOS sent but never reaching the sinks, still fails after eosTimeout exactly as before.

The new controller test reproduces the scenario end-to-end with a real pipeline (silence generator -> opus -> ogg file sink): it sends EOS directly through the pipeline, invokes SendEOS the moment the bus EOS is processed, and asserts the egress completes with a finalized file. Without the guard it fails deterministically with the production signature, EGRESS_FAILED (error: "pipeline frozen").

### DIFF
--- a/pkg/pipeline/controller_test.go
+++ b/pkg/pipeline/controller_test.go
@@ -120,7 +120,7 @@ func TestUnsolicitedEOSBeforeSendEOS(t *testing.T) {
 
 	// As soon as the bus EOS is processed (before Close marks the egress
 	// complete), invoke the normal end path — mirroring the source-side
-	// shutdown that raced the unsolicited EOS in production.
+	// shutdown path.
 	sendEOSDone := make(chan struct{})
 	go func() {
 		defer close(sendEOSDone)
--- a/pkg/pipeline/controller_test.go
+++ b/pkg/pipeline/controller_test.go
@@ -1,0 +1,168 @@
+// Copyright 2026 LiveKit, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


paulwe (approved):

pkg/pipeline/controller_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+	select {
+	case <-c.playing.Watch():
+	case <-time.After(10 * time.Second):
+		t.Fatal("pipeline did not reach PLAYING")
+	}
+
+	// As soon as the bus EOS is processed (before Close marks the egress
+	// complete), invoke the normal end path — mirroring the source-side
+	// shutdown that raced the unsolicited EOS in production.
```
boks1971: have been telling Claude to not do references like this. `in production` does not mean anything for somebody reading this a couple of months from now.

boks1971 (approved):

PR #1316, opened 2026-07-31
Add videotestsrc delay headroom to audio_queue to stop HLS audio drops

milos-lk:
## Problem

Participant egress with segments (HLS) output steadily discards audio in the leaky `audio_queue` whenever the participant publishes audio but no video. Reported case: 1,731 buffers ≈ 33% of all audio missing from a 112s recording, gaps throughout. participant+file and room_composite are unaffected.

## Root cause

With no video track, the splitmuxsink reference stream is the black `videotestsrc` branch, which is delayed 2s by `min-threshold-time` on `video_test_src_queue` (added in 84d9654 / #725 to keep the test-src PTS watermark behind a late-arriving real track, so the selector monotonicity probes don't drop the camera's first frames — the 2s matches the jitter buffer latency).

splitmuxsink admits an audio buffer with PTS `t` only after a reference (video) buffer with PTS > `t` has arrived (`gstsplitmuxsink.c` non-reference gate). With the reference 2s + encoder-lookahead late, every audio buffer waits ~2.2s of the leaky `audio_queue`'s 3s (`PipelineLatency`) budget at steady state. Fragment closes stall the reference further; the queue tops out and `leaky=downstream` silently discards the oldest buffers — each one a hole in the muxed audio.

## Fix

Size `audio_queue` at `PipelineLatency + videoTestSrcDelay` for SDK pipelines with video enabled — i.e. exactly when the test-src branch can be the mux reference — restoring the headroom the delay consumes. The 2s literal is promoted to a shared `videoTestSrcDelay` constant so the delay and the audio headroom can't drift apart. The queue stays leaky so a genuinely wedged downstream still degrades gracefully instead of stalling the audio path.

No latency or memory impact: the cap is not a target (occupancy is set by the muxer gate either way), and the queue holds encoded audio (~32KB per extra second).

## Testing

New segments integration test `ParticipantComposite/AudioOnlyPublisher` (audio-only publisher, 6s segments, mirrors the reported production request). Before this change it fails via the beep-presence content check:

- repro: `droppedByQueue: {audio_queue: 289}` (~6.2s of audio), beeps 26/30 (87%) → FAIL
- control (`ParticipantComposite/VP8`, same config + camera): 0 dropped, beeps 30/30 → PASS

With the change the repro passes alongside the existing segments suite.

### DIFF
--- a/pkg/pipeline/builder/video.go
+++ b/pkg/pipeline/builder/video.go
@@ -523,7 +523,7 @@ func (b *VideoBin) addVideoTestSrcBin() error {
 	if err != nil {
 		return err
 	}
-	if err = queue.SetProperty("min-threshold-time", uint64(videoTestSrcDelay)); err != nil {
+	if err = queue.SetProperty("min-threshold-time", uint64(videoTestSrcDelay.Nanoseconds())); err != nil {
 		return errors.ErrGstPipelineError(err)
 	}
 
--- a/pkg/pipeline/builder/video.go
+++ b/pkg/pipeline/builder/video.go
@@ -31,7 +31,8 @@ import (
 )
 
 const (
-	videoTestSrcName = "video_test_src"
+	videoTestSrcName  = "video_test_src"
+	videoTestSrcDelay = 2 * time.Second
 )
 
 type VideoBin struct {
@@ -522,7 +523,7 @@ func (b *VideoBin) addVideoTestSrcBin() error {
 	if err != nil {
 		return err
 	}
-	if err = queue.SetProperty("min-threshold-time", uint64(2e9)); err != nil {
+	if err = queue.SetProperty("min-threshold-time", uint64(videoTestSrcDelay)); err != nil {
 		return errors.ErrGstPipelineError(err)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/pipeline/builder/video.go
```diff
@@ -522,7 +523,7 @@ func (b *VideoBin) addVideoTestSrcBin() error {
 	if err != nil {
 		return err
 	}
-	if err = queue.SetProperty("min-threshold-time", uint64(2e9)); err != nil {
+	if err = queue.SetProperty("min-threshold-time", uint64(videoTestSrcDelay)); err != nil {
```
boks1971: :nit: just my personal preference as it makes my head hurt less to have to remember that time.Duration is actually int64, can do `videoTestSrcDelay.Nanoseconds()` and it makes it clear the units are nano seconds.

boks1971 (approved):

pkg/pipeline/builder/video.go
milos-lk: Actually that would break it here - Nanoseconds() returns int64, so go-glib builds a G_TYPE_INT64 GValue. The queue's min-threshold-time property is guint64, and go-glib's SetPropertyValue does a strict type equality check - no implicit conversion. Will leave it as is (def clearer than 2e9)

boks1971: Sounds good.

No need to change. But I was thinking the cat can stay, but remove the cognitive load of time.Duration is just nanoseconds.

milos-lk: ah got it now - that's good - we can do it

milos-lk: updated

PR #1325, opened 2026-08-03
Separate admission costs for SDK-source audio room composites

milos-lk:
## Motivation

Audio-only room composites without a `layout` or `custom_base_url` run on the SDK source (no chrome) and use a small fraction of the CPU and memory of the chrome path sharing their cost class. Since admission costs also bound how quickly a node can absorb new requests during the hold window, pricing the SDK path separately allows much denser packing and faster ramp for audio-only workloads, without touching the chrome-based costs.

## Changes

**Commit 1 — refactor**: extracts the duplicated cost selection in `canAcceptRequestLocked` and `AcceptRequest` into a single `costsForRequest` helper returning `{cpu, memory, isWeb}`, parameterizes the memory admission check, and adds a pricing table test covering all v1/v2 request shapes. Also fixes v2 template/web requests not being rejected when the pulse client limit is reached (the closure's rejection result was ignored at its call sites, so only v1 requests were enforced).

Preserved behavior (now visible in the table test): v2 `Media` requests are held at `ParticipantCpuCost` regardless of kind, and a v2 request with no recognizable source costs 0.

**Commit 2 — the split**: adds `sdk_audio_room_composite_cpu_cost` and `sdk_audio_room_composite_memory_cost`, applied when `ShouldUseSDKSource` is true for audio-only room composites (v1) and audio-only templates (v2, including replay) — the same predicate the pipeline uses for source selection, so admission cost and the actual pipeline can't disagree.

Both default to `audio_room_composite_cpu_cost` / `memory_cost` when unset, so **behavior is unchanged until explicitly configured**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ backup_storage: files will be moved here when uploads fail. location must have w
 enable_chrome_sandbox: if true, egress will run Chrome with sandboxing enabled. This requires a specific Docker setup, see below.
 cpu_cost: # optionally override cpu cost estimation, used when accepting or denying requests
   room_composite_cpu_cost: 3.0
+  audio_room_composite_cpu_cost: 1.0
+  sdk_audio_room_composite_cpu_cost: 0.5 # audio-only room composites without layout or custom_base_url (no chrome). Defaults to audio_room_composite_cpu_cost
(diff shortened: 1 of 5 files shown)


boks1971 (approved):
nice!

biglittlebigben (approved):

PR #1326, opened 2026-08-03
Support dual-channel audio mixing in web-sourced room composites

### LINKED ISSUE #1296: [FEATURE] DUAL_CHANNEL_AGENT supports upload of video file with dual channel audio
**Is your feature request related to a problem? Please describe.**

To get dual channel audio recording with a composite room recording (or recording of participant alone) we currently need to launch two seperate egress requests and manually syncronise them. 

```
        req = api.RoomCompositeEgressRequest(
            room_name=self.ctx.room.name,
            audio_only=True,
            audio_mixing=api.AudioMixing.DUAL_CHANNEL_AGENT,
            file=api.EncodedFileOutput(file_type="OGG", filepath=filepath, s3=s3_upload),
        )
```
and
```
        req = api.RoomCompositeEgressRequest(
            room_name=self.ctx.room.name,
            layout="grid",
            preset=api.EncodingOptionsPreset.H264_720P_30,
            audio_only=False,
            file=api.EncodedFileOutput(file_type="MP4", filepath=filepath, s3=s3_upload),
        )
```

We do this to be able to have a recording of the call with an AI agent, alongside audio of the user alone that we can post-process (not contaminated by agent audio). We need to align the outputs to the video in our business case (i.e. to show when things were said in the video). 

**Describe the solution you'd like**
Ability to perform a single egress request that results in a video file with composite room recording and dual channel audio (split by agent/participant). 

Like a lot of users of AI agents, we are not very interested in recording the video of the AI agent so in principle rather than full room composite recording, just video recording of the human participant, and dual channel (split by agent/human user) audio woud be sufficient. 

 

**Describe alternatives you've considered**

We have considered recording audio directly from the agent by track, and uploading that, then syncronising, but this adds local CPU load and complexity we would prefer to avoid. Our current workaround is to do two egress requests, but this is problematic in that during peak load we can hit very high egress concurrency and is not optimally efficient.

milos-lk:
Fixes livekit/egress#1296

`DUAL_CHANNEL_AGENT` and `DUAL_CHANNEL_ALTERNATE` previously only applied to SDK-sourced audio-only room composites, since Chrome mixes all participants into a single stereo stream before egress captures it from the pulse sink. Customers wanting a video recording alongside agent/user-separated audio had to run two egresses and synchronize the outputs themselves.

The split now happens inside the template, before Chrome's mixdown — the only point in the web path where per-participant audio still exists. When a room composite requests dual-channel mixing, egress appends `audioMixing=<mode>` to the template URL, and a new `DualChannelAudioRenderer` replaces `RoomAudioRenderer`: each subscribed track plays through a WebAudio `StereoPannerNode`, agents hard left (`participant.isAgent`) and everyone else hard right, or alternating per track with assignments pinned for the track's lifetime. Chrome's mix then carries channel-separated audio that stays time-aligned with the video by construction, with no GStreamer pipeline changes. Recordings without the option behave exactly as before. The mode is also exposed as `EgressHelper.getAudioMixing()` in the template SDK (bumped to 0.2.2) so custom templates can honor it the same way.

The integration test records a grid composite with an agent participant and verifies every detected beep lands on its expected channel, alongside the standard presence/cadence/AV-sync checks. Getting this to pass required a detector fix in media-samples ([livekit/media-samples#15](<https://github.com/livekit/media-samples/issues/15>)): bandpass bleed from neighboring tone frequencies was winning the detection race under Chrome's per-track playout skew and misattributing channels on recordings that were in fact correctly separated.

Follow-ups after this merges: re-pin `go.mod` from the media-samples branch to its merge commit, tag `template0.2.2` to publish the SDK, then switch the default template from the direct query-string read to `EgressHelper.getAudioMixing()`. Adding `audio_mixing` to the v2 `TemplateSource` proto is a separate protocol change. The template publish workflow will pin the new template image tag on this branch automatically.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


boks1971 (approved):

biglittlebigben (approved):

biglittlebigben:
@frostbyte73 could you have a look at this as well?

milos-lk:
thanks @frostbyte73 for the offline feedback - the v1 channel routing settings should sunset in favor of v2 - I will close the PR and start preparing v2 variant

PR #1338, opened 2026-08-11
Raise minimum image upload queue capacity to absorb frame bursts

milos-lk:
Follow-up to #1336, which made `ImageSink.NewImage` a non-blocking send that fails the egress when the upload queue is full. The queue capacity is computed as `max_upload_queue * 60 / capture_interval`, so a capture interval of 3600s or longer yields at most one slot. That single slot is overflowed by a normal startup frame burst — `videorate`'s skip-to-first frame and the first rate-aligned frame arrive back-to-back while the first upload is still in flight — and the egress fails immediately with `image upload queue full (1 pending)`. Caught on a real hourly-snapshot egress right after #1336 rolled out.

The capacity computation moves into `imageQueueCapacity` with a floor of `minPendingUploads = 4`: enough for every burst source we can name (startup burst, EOS flush) with margin, while a full queue still means a genuine multi-interval upload backlog. Queue slots are two-field structs, so the floor costs nothing. Includes a regression test pinning the 3600s case and the formula boundaries.

### DIFF
--- a/pkg/pipeline/sink/image.go
+++ b/pkg/pipeline/sink/image.go
@@ -54,6 +54,18 @@ type imageUpdate struct {
 	filename  string
 }
 
+// covers frame bursts that outpace the capture interval: videorate's
+// skip-to-first plus the first rate-aligned frame at startup, and the EOS flush
+const minPendingUploads = 4
+
+func imageQueueCapacity(maxUploadQueue int, captureInterval uint32) int {
+	capacity := (maxUploadQueue * 60) / int(captureInterval)
+	if capacity < minPendingUploads {
+		capacity = minPendingUploads
+	}
+	return capacity
+}
+
 func newImageSink(
 	p *gstreamer.Pipeline,
 	conf *config.PipelineConfig,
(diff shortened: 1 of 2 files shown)


pkg/pipeline/sink/image.go
```diff
@@ -54,6 +54,18 @@ type imageUpdate struct {
...
+// covers frame bursts that outpace the capture interval: videorate's
+// skip-to-first plus the first rate-aligned frame at startup, and the EOS flush
+const minPendingUploads = 4
+
+func imageQueueCapacity(maxUploadQueue int, captureInterval uint32) int {
+	capacity := (maxUploadQueue * 60) / int(captureInterval)
+	if capacity < minPendingUploads {
+		capacity = minPendingUploads
+	}
+	return capacity
```
boks1971: `return max((maxUploadQueue * 60) / int(captureInterval), minPendingUploads)` maybe.

boks1971 (approved):

pkg/pipeline/sink/image_test.go
```diff
@@ -162,3 +162,13 @@ func TestImageSinkCloseDrainsPendingUploads(t *testing.T) {
 	default:
 	}
 }
+
+// Regression: a capture interval of exactly 3600s used to yield a single-slot
+// queue, which the startup frame burst overflows, failing the egress.
```
boks1971: These comments using `Regression` is confusing again.

milos-lk: need to make claude respect these better - updated my CLAUDE.md with a bit more explicit rule on this.

PR #1342, opened 2026-08-15
egress: implement v2 passthrough, reach TrackEgress parity

frostbyte73:
A v2 livekit.StartEgressRequest that names one track (MediaSource.video_track_id, or a single AudioRoute matching track_id) with passthrough set and a DEFAULT_FILETYPE FileOutput behaves exactly like v1 StartTrackEgress: no decode/encode, file extension derived from the track codec, track-based filename replacements, single-track admission cost, websocket/video rejection.


pkg/pipeline/controller.go
```diff
@@ -424,37 +424,6 @@ func (c *Controller) UpdateStream(ctx context.Context, req *livekit.UpdateStream
 	return errs.ToError()
 }
 
-func (c *Controller) UpdateEgress(ctx context.Context, req *livekit.UpdateEgressRequest) error {
```
biglittlebigben, line 427: I was under the impression that UpdateEgress always failed. However since we have this alias for UpdateStream, we cannot remove this unless we can prove that nobody uses this API.

biglittlebigben (commented):
Can you respond to the Devin comments?

pkg/pipeline/controller.go
frostbyte73: it was never released in protocol

frostbyte73:
> Can you respond to the Devin comments?

yeah, more fixes/updates coming

biglittlebigben (approved):

pkg/config/pipeline.go
```diff
@@ -692,6 +717,57 @@ func (p *PipelineConfig) applyV2Source(req egress.EgressRequest) (connectionInfo
...
+		case len(media.Audio.Routes) == 1:
+			m, ok := media.Audio.Routes[0].Match.(*livekit.AudioRoute_TrackId)
+			if !ok {
+				return errors.ErrInvalidInput("passthrough audio route match")
+			}
+			audioTrackID = m.TrackId
+		}
+	}
+
+	if (videoTrackID == "") == (audioTrackID == "") {
```
milos-lk, line 757: Should the empty-string case be rejected here? validatePassthrough reads presence from the string value, but applyV2Source reads it from oneof/slice presence, so video_track_id: "TR_vid" plus a route with track_id: "" passes this check as "audio absent", then sets AudioEnabled = true and overwrites p.TrackID with "". Worth requiring both ids to be non-empty?

pkg/config/pipeline.go
```diff
@@ -662,7 +684,7 @@ func (p *PipelineConfig) applyV2Source(req egress.EgressRequest) (connectionInfo
 				p.CaptureAudioAll = true
 			} else if len(media.Audio.Routes) > 0 {
 				p.AudioEnabled = true
-				p.AudioTranscoding = true
+				p.AudioTranscoding = !p.Passthrough
```
milos-lk, line 687: wondering if this flag is worth keeping? Don't see the case where it ever gets false value

pkg/pipeline/source/track_worker.go
```diff
@@ -238,14 +238,7 @@ func (s *SDKSource) updatePreInitStateLocked(op Operation, ts *config.TrackSourc
...
-	switch s.RequestType {
-	case types.RequestTypeTrackComposite:
-		if s.Identity == "" || track.Kind() == webrtc.RTPCodecTypeVideo {
-			s.Identity = rp.Identity()
-			s.filenameReplacements["{publisher_identity}"] = s.Identity
-		}
-
-	case types.RequestTypeTrack:
+	if s.Passthrough {
 		s.Identity = rp.Identity()
```
milos-lk, line 242: This now runs for RequestTypeMedia, which is the one request type that goes through awaitMediaTracks and that reads s.Identity unsynchronized while this writes it under s.mu.  Should the read side take the lock?

pkg/config/pipeline.go
frostbyte73: it's not really used today but might as well keep it, we might end up using it with replay

milos-lk (approved):

PR #1348, opened 2026-08-17
fix: strip packet trailers from VP9 layer frames before decode

### LINKED ISSUE #1347: VP9 SVC recordings corrupt when the publisher uses packet trailers / frame metadata (per-layer trailers reach the decoder)
## Summary

When a publisher enables the SDK packet trailer / frame metadata feature (`packet_trailer_features` on rust-sdks ≤ the 1.1.11-era API, `frame_metadata_features` after the rename) and publishes VP9 at resolutions where the SDK's default quality layers apply, **every recording of that track is corrupt**: `vp9dec` logs `corrupt frame` on essentially every frame, the job ends with `GstVP9Dec: No valid frames decoded before end of stream`, and the output is an empty/aborted recording.

## Mechanism

1. The FFI's sender frame transform appends the LKTS trailer **per encoded frame**. With VP9 SVC (how the SDKs deliver VP9 quality layers — `MULTIPLE_SPATIAL_LAYERS_PER_STREAM`), libwebrtc emits **each spatial layer as its own encoded frame**, so a 3-layer publish carries three trailers per picture.
2. livekit-server's per-subscriber `StripTrailer` truncates only the **frame tail** (marker packet). The lower layers' trailers stay embedded **mid-superframe**, where no tail-strip can reach them.
3. Egress feeds raw RTP to GStreamer (`appsrc → rtpvp9depay → vp9dec`) with a `NullTranslator` for VP9 — no trailer handling exists in egress at any version including main. After depayload, the reassembled superframe reaches libvpx, which tolerates only **zero** bytes between frames (`vp9_dx_iface.c`, "Account for suboptimal termination"). The trailer's first byte is always `0x01 ^ 0xFF = 0xFE` (the XOR-0xFF design that protects H264 start codes), so libvpx parses it as a bogus frame header → `UNSUP_BITSTREAM`, every frame.

Why only VP9: H264 tolerates trailing bytes inside NAL framing; VP8 has no spatial SVC (single layer — the SFU's tail strip covers its one trailer); AV1 gets spec-legal metadata OBU carriage (`av1::InsertTrailerObu`). VP9 has no legal carriage point at all — raw concatenation is invalid bitstream by construction, and the anti-H264-collision XOR guarantees the bytes are never the zeros libvpx would forgive.

## Reproduction

- Publisher: any rust-sdks-based SDK with the trailer feature armed and frames carrying metadata (e.g. python `livekit` wheel, `frame_metadata_features=[FMF_USER_TIMESTAMP]` + `capture_frame(..., metadata=FrameMetadata(user_timestamp=...))`), VP9, ≥ ~640x360 so the encoder auto-creates 2–3 spatial layers.
- Record the participant with ParticipantEgress (file output).
- Result: decoder errors on every frame; no usable recording. Trailer-off or `scalability_mode="L1T3"` publishes record cleanly (controls).

Reproduced against: livekit-server v1.11.0 and v1.13.5 (cap-gated strip), egress v1.13.0, v1.14.0, and current main (`f0780cc`), GStreamer 1.24.12 and 1.26.7. No SFU/egress/GStreamer version combination avoids it — the mid-superframe trailers are unreachable by design for every existing strip.

## Impact

In our production (recordings enabled fleet-wide, trailer usage ramping on the client side), this produced hundreds of thousands of corrupt recordings per week and drove egress job failure rates of ~16–20%; after deploying the fix below as a fork, corruption went to zero at the rollout boundary and the failure rate dropped to ~2%.

## Fix

PR incoming: a `VP9Translator` in the appwriter's existing per-packet translator hook that checks end-of-layer-frame (E-bit) packets for the LKTS envelope and truncates it. The RTP layer is the only point where layer-frame boundaries still exist — post-depayload, VP9 frames self-delimit only through the entropy decoder, which is exactly where they currently explode. Trailer-less payloads pass through byte-identically; includes table tests for the envelope edge cases.

An alternative/complementary publisher-side fix — appending the trailer only to the **final** spatial layer of each picture, so the existing SFU tail-strip suffices — would address the class at the source (rust-sdks), but egress-side stripping is needed anyway for publishers already in the wild.

ilay-decart:
Fixes #1347.

## What

Adds a `VP9Translator` in the appwriter's existing per-packet translator hook (the same seam `VP8Translator` uses) that, on end-of-layer-frame packets (VP9 payload descriptor E bit), checks the payload tail for the packet-trailer envelope (`[len^0xFF]["LKTS"]`) and truncates it. VP9 previously used `NullTranslator`; all other codecs are untouched.

## Why

Full mechanism in #1347. Short version: the SDK sender transform appends the trailer per encoded frame; VP9 SVC makes each spatial layer its own encoded frame, so lower-layer trailers end up mid-superframe where the server's tail-only strip can't reach them; after `rtpvp9depay` reassembly libvpx rejects every frame (non-zero inter-frame bytes are invalid VP9, and the trailer's XOR-0xFF design guarantees a non-zero first byte). The RTP layer is the last point where layer boundaries exist, which is why the strip lives in the translator hook rather than anywhere downstream.

## Safety

- Strips only when all four hold: payload parses as a VP9 descriptor, E bit set, tail is the `LKTS` magic, and the length byte is coherent for the post-descriptor payload. Trailer-less payloads pass through byte-identically.
- Already-stripped tails (the server handles the marker packet for cap-less subscribers) no-op — the strip is idempotent.
- Residual false-positive requires legitimate video whose E-packet tail spells the 4-byte magic with a coherent length byte (~2^-32 per packet); documented in the tests.

## Testing

- `translator_vp9_test.go`: table tests covering the three real trailer formats (15B timestamp, 21B +frame_id, 255B max user_data with its 0x00 length byte), V=1/F=1 descriptor shapes, exact-fit boundary (`trailerLen == len(payload)`), descriptor-overlap magic windows, malformed length bytes, marker-without-E, padding-like payloads, idempotency, and the byte-prefix property.
- End-to-end against this branch's build (real SFU + rust-sdks publisher, participant egress): stock main reproduces #1347 (decoder errors on every frame, aborted job at 3-spatial-layer 1280x704); this branch records clean, full-length, fully decodable MP4s at 2- and 3-layer SVC across the trailer formats above, with trailer-off / H264 / VP8 / L1T3 controls unchanged.
- The same change (applied to v1.13.0) has been running across our production recording fleets: recording corruption went from hundreds of thousands of files/week to zero at the rollout boundary, and egress job failure rate dropped ~16–20% → ~2%, stable since.

### DIFF
--- a/pkg/pipeline/source/sdk/appwriter.go
+++ b/pkg/pipeline/source/sdk/appwriter.go
@@ -204,7 +204,7 @@ func NewAppWriter(
 
 	case types.MimeTypeVP9:
 		depacketizer = &codecs.VP9Packet{}
-		w.translator = NewNullTranslator()
+		w.translator = NewVP9Translator()
 
 	default:
 		return nil, errors.ErrNotSupported(string(ts.MimeType))
--- a/pkg/pipeline/source/sdk/translator.go
+++ b/pkg/pipeline/source/sdk/translator.go
@@ -15,9 +15,11 @@
 package sdk
 
 import (
+	"bytes"
 	"time"
 
 	"github.com/pion/rtp"
+	"github.com/pion/rtp/codecs"
 
 	"github.com/livekit/livekit-server/pkg/sfu/buffer"
 	"github.com/livekit/livekit-server/pkg/sfu/codecmunger"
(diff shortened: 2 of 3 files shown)


cnderrauber:
Fixed by sfu https://github.com/livekit/livekit/pull/4773. Close this as there is no feedback. Feel free to reopen this if you still have an issue.

PR #1353, opened 2026-08-20
Bound websocket sink writes with a write deadline

milos-lk:
## Motivation

A websocket track egress whose consumer stops reading currently buffers unboundedly instead of failing:

- `WebsocketSink.Write` calls `conn.WriteMessage` with no deadline from inside the appsink `NewSampleFunc` callback, so a stalled TCP peer blocks the callback indefinitely.
- ~The appsink is created without a `max-buffers` bound (`builder.BuildWebsocketBin`), so while the callback is blocked, every incoming sample queues in the appsink.~
- The ping goroutine's `WriteMessage` can block on the same stalled connection **while holding the sink mutex**, and `Close` can block writing the close message — so even shutdown stalls.

Observed in production — a 3h05m websocket audio stream to a consumer behind a CDN that read slowly and eventually reset. Peak egress memory reached **2.48 GB** — consistent with hours of raw PCM at ~190 KB/s accumulating in the appsink queue — before the peer's RST finally failed the egress with `write: broken pipe`.

## Changes

All websocket writes (samples, ping, pong, mute/unmute text, close) go through a `writeMessageLocked` helper that sets a **30s write deadline** before each write. A consumer that stops reading for 30s now fails the sample write, which propagates through the existing `NewSampleFunc` error path as a destination error (`psrpc.Unavailable`) and ends the egress — bounding worst-case buffering to ~30s of media (a few MB for audio) instead of hours.

30s is deliberately generous: it only triggers when the peer has accepted no TCP data at all for the full window, i.e. the consumer is effectively gone — transient slow reads that drain within the window are unaffected (a live stream 30s behind real time is already unrecoverable for the reader).

### DIFF
--- a/pkg/pipeline/sink/websocket.go
+++ b/pkg/pipeline/sink/websocket.go
@@ -36,7 +36,14 @@ import (
 	"github.com/livekit/psrpc"
 )
 
-const pingPeriod = time.Second * 30
+const (
+	pingPeriod = time.Second * 30
+
+	// writeTimeout bounds every websocket write; without it, a consumer that
+	// stops reading blocks the appsink callback and samples accumulate in the
+	// unbounded appsink queue for as long as the connection stays up.
+	writeTimeout = time.Second * 30
+)
 
 type WebsocketSink struct {
 	*base
@@ -125,7 +132,7 @@ func (s *WebsocketSink) Start() error {
 		s.mu.Lock()
 		defer s.mu.Unlock()
 
-		_ = s.conn.WriteMessage(websocket.PongMessage, []byte("pong"))
+		_ = s.writeMessageLocked(websocket.PongMessage, []byte("pong"))
 		return nil
 	})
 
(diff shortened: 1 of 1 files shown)


boks1971 (approved):

biglittlebigben (approved):
Orthogonal to this change, but how does psrpc.Unavailable get translated by the time it gets to the client? By default psrpc would translate this to 503. Maybe 408 would be more appropriate here?

milos-lk:
> Orthogonal to this change, but how does psrpc.Unavailable get translated by the time it gets to the client? By default psrpc would translate this to 503. Maybe 408 would be more appropriate here?

Good point - it's going to be 503.
Not sure if 408 is the semantically the best code, also psrpc's table has no 408 at all. Maybe better `424 Failed Dependency` - that's probably more honest fit? Will think about that (wrt retries and consistency) and eventually make that change in an independent PR.

PR #1358, opened 2026-08-28
Send EOS for every appsrc linked into the pipeline (CS-1547)

pksgit:
Fixes [CS-1547](https://linear.app/livekit/issue/CS-1547/pipeline-frozen-when-writer-added-during-closewriters-race-window). Related customer report: CSCU-18.

## The problem

A track subscribed during the `CloseWriters()` window has its appsrc linked into the running pipeline by `OnTrackAdded()`, but never sends EOS. Cleanup in `AppWriter.start()` guards on `playing.IsBroken()`, and the PLAYING notification that would break that fuse is dropped by `submitOp()` once `closing` is set — permanently, not late. The mixer waits on that input pad forever: `endStreamProcessed not broken after 3 seconds`, then `ErrPipelineFrozen` 30s later. The recording is lost, not delayed: `Close()` skips `uploadManifest()` on `EGRESS_FAILED`.

## The fix

Guard cleanup on a new `addedToPipeline` fuse, broken where the appsrc actually becomes part of the pipeline:

- Marked once in `handleSubscribe`, before the appsrc is linked — by `OnTrackAdded()` for post-init subscriptions, or by `BuildPipeline()` for tracks present at startup. Marking after the link would leave a window where the appsrc is in the pipeline but the writer, on its own goroutine, does not yet know it owes an EOS. Over-marking costs a no-op `EndStream()`; under-marking freezes a recording.
- Marking before the link also covers a startup track that ends during the build: its source is already in the pipeline config, so `BuildPipeline` links the appsrc, and passthrough / track composite never remove that source bin. An EOS sent before the element is linked is retained and delivered once the pipeline starts it.
- `Playing()` breaks the fuse too, since reaching PLAYING implies being linked.

`onEOSSent()` also returns early while `BuildReady` is open. A track that EOFs mid-build otherwise reaches `c.p.Stop()` before `BuildPipeline()` has assigned `c.p` — a nil dereference and a race with that write. This is what made the earlier attempt (cfce5d0) panic in production.

## Verifying in production

Cleanup logs a warning when it sends EOS for a writer that was never reported PLAYING — the race happening and being handled:

```
appsrc never reported PLAYING, sending EOS anyway
  trackID=… kind=… active=… draining=… unsubscribed=… lastReceived=…
```

Before this change the same egress logged `endStreamProcessed not broken after 3 seconds` and failed. So `service:cloud-egress "appsrc never reported PLAYING"` counts races caught, and an egress logging both that and `endStreamProcessed not broken` would mean the fix was insufficient — worth alerting on. Expect it to be rare and confined to shutdown.

## Tests

| Test | Pins | Without the fix |
| --- | --- | --- |
| `TestShouldSendEOS` | The cleanup decision across three states: linked but never reported PLAYING must still owe EOS. | fails |
| `TestOnEOSSentBeforePipelineBuilt` | Cleanup reaching `onEOSSent()` mid-build must not dereference `c.p` or start the EOS sequence. | panics |
| `TestPlayingLostAfterCloseWriters` / `…DeliveredBeforeClose` | `OpPlaying` is dropped once `closing` is set, `OpClose` is not. | n/a |

No integration test: the dynamic-add path is already covered end to end by `testRoomCompositeLateTrackDuration`, and a test targeting the race itself would have to land a stop inside a few-millisecond window, so it would be green whenever it failed to reproduce. A deterministic version behind a test-only hook in `sdk_testing.go` is doable if wanted.

## Unrelated, found while testing

Both confirmed on `main` at feae0c7, not touched here:

- `Controller.Close()` racing `EgressInfo.UpdateStatus()` from `SendEOS()` on `c.Info` — both production paths, surfaced by `TestUnsolicitedEOSBeforeSendEOS`.
- `go test ./pkg/pipeline/ -count>1` panics on duplicate Prometheus collector registration, so repeat `-race` runs need `-count=1` in a loop.
- `TestLivePlaylistWriter` (`sink/m3u8`) and `TestUploader` (`sink/uploader`, needs AWS credentials) also fail on `main`.

### DIFF
--- a/pkg/pipeline/source/sdk/appwriter.go
+++ b/pkg/pipeline/source/sdk/appwriter.go
@@ -273,7 +273,7 @@ func (w *AppWriter) start() {
 	if w.shouldSendEOS() {
 		if !w.playing.IsBroken() {
 			// Linked into the pipeline but never reported PLAYING: the CloseWriters
-			// race (CS-1547). Expect this to be rare and confined to shutdown.
+			// race. Expect this to be rare and confined to shutdown.
 			w.logger.Warnw("appsrc never reported PLAYING, sending EOS anyway", nil,
 				"active", w.active.Load(),
 				"draining", w.draining.IsBroken(),
@@ -693,7 +693,7 @@ func (w *AppWriter) MarkAddedToPipeline() {
 //
 // The test is whether the appsrc is linked into the pipeline, not whether it was
 // reported PLAYING: once CloseWriters() sets closing that notification is dropped,
-// while the pipeline still waits on the pad (CS-1547).
+// while the pipeline still waits on the pad.
 func (w *AppWriter) shouldSendEOS() bool {
 	return w.addedToPipeline.IsBroken()
 }
--- a/pkg/pipeline/source/sdk/appwriter.go
+++ b/pkg/pipeline/source/sdk/appwriter.go
@@ -104,6 +104,7 @@ type AppWriter struct {
 	lastReceived             atomic.Time
 	lastPushed               atomic.Time
 	playing                  core.Fuse
+	addedToPipeline          core.Fuse
 	draining                 core.Fuse
 	unsubscribed             core.Fuse
 	endStreamSignaled        core.Fuse
@@ -260,6 +261,7 @@ func (w *AppWriter) start() {
 		w.logger.Errorw("endStreamProcessed not broken after 3 seconds, bug in the draining logic!", nil,
 			"endStreamSourceProcessed", w.endStreamSourceProcessed.IsBroken(),
 			"playing", w.playing.IsBroken(),
+			"addedToPipeline", w.addedToPipeline.IsBroken(),
 			"active", w.active.Load(),
 			"lastReceived", w.lastReceived.Load(),
 			"lastPushed", w.lastPushed.Load(),
@@ -268,7 +270,17 @@ func (w *AppWriter) start() {
 	}
 
 	// clean up
-	if w.playing.IsBroken() {
+	if w.shouldSendEOS() {
+		if !w.playing.IsBroken() {
+			// Linked into the pipeline but never reported PLAYING: the CloseWriters
+			// race (CS-1547). Expect this to be rare and confined to shutdown.
+			w.logger.Warnw("appsrc never reported PLAYING, sending EOS anyway", nil,
+				"active", w.active.Load(),
+				"draining", w.draining.IsBroken(),
+				"unsubscribed", w.unsubscribed.IsBroken(),
+				"lastReceived", w.lastReceived.Load(),
+			)
+		}
 		w.callbacks.OnEOSSent()
 		flow := w.src.EndStream()
 		if flow == gst.FlowFlushing {
@@ -665,7 +677,25 @@ func (w *AppWriter) maybeCheckPipelineLag(pts time.Duration) {
 }
 
 func (w *AppWriter) Playing() {
+	// Reaching PLAYING implies the appsrc is linked. handleSubscribe normally
+	// marks it first; this keeps the invariant if a marking call is ever missed.
+	w.addedToPipeline.Break()
 	w.playing.Break()
+}
+
+// MarkAddedToPipeline signals that the appsrc is linked into the pipeline, which
+// blocks EOS aggregation on this writer until it ends its stream.
+func (w *AppWriter) MarkAddedToPipeline() {
+	w.addedToPipeline.Break()
+}
+
+// shouldSendEOS reports whether cleanup must push EOS into this writer's appsrc.
+//
+// The test is whether the appsrc is linked into the pipeline, not whether it was
+// reported PLAYING: once CloseWriters() sets closing that notification is dropped,
+// while the pipeline still waits on the pad (CS-1547).
+func (w *AppWriter) shouldSendEOS() bool {
+	return w.addedToPipeline.IsBroken()
 }
 
 // Drain blocks until finished
--- a/pkg/pipeline/source/sdk/appwriter.go
+++ b/pkg/pipeline/source/sdk/appwriter.go
@@ -272,8 +272,8 @@ func (w *AppWriter) start() {
 	// clean up
 	if w.shouldSendEOS() {
 		if !w.playing.IsBroken() {
-			// Linked into the pipeline but never reported PLAYING: the CloseWriters
-			// race. Expect this to be rare and confined to shutdown.
+			// Linked into the pipeline but never reported PLAYING. Expect this to be
+			// rare and confined to shutdown.
 			w.logger.Warnw("appsrc never reported PLAYING, sending EOS anyway", nil,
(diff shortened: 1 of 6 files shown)


pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -268,7 +270,22 @@ func (w *AppWriter) start() {
 	}
 
 	// clean up
-	if w.playing.IsBroken() {
+	if w.shouldSendEOS() {
+		if !w.playing.IsBroken() {
+			// The appsrc was linked into the pipeline but its PLAYING notification
+			// never arrived - the CloseWriters race (CS-1547). Under the old
```
biglittlebigben: Please update this comment. Full history is not needed in the code comment, only an explanation of the code as it currently is.

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -665,9 +682,40 @@ func (w *AppWriter) maybeCheckPipelineLag(pts time.Duration) {
 }
 
 func (w *AppWriter) Playing() {
+	// Backstop, not the primary signal: reaching PLAYING implies the appsrc is
```
biglittlebigben: Please fix this comment to be more succinct.

biglittlebigben (commented):
Please adjust the comments to be better scoped and only describe the current implementation

pkg/pipeline/source/sdk/appwriter.go
pksgit: Yeah I was going to make some updates including the comments, before making this ready for review. This is still in draft mode. The initial comments from Claude are sometimes too verbose and include some unnecessary details from the prompt session.

biglittlebigben: I think most folks have claude instructions in their conf to request shorter, descriptive comments

pkg/pipeline/source/sdk/appwriter_test.go
```diff
@@ -0,0 +1,243 @@
...
+	})
+
+	t.Run("added to the pipeline but PLAYING never delivered", func(t *testing.T) {
+		w := &AppWriter{}
+		w.MarkAddedToPipeline()
+
+		require.False(t, w.playing.IsBroken(),
+			"precondition: this is the race - the PLAYING notification was dropped")
+		require.True(t, w.shouldSendEOS(),
+			"CS-1547: the appsrc is linked into the pipeline, so cleanup must send EOS "+
```
milos-lk: nit: I would suggest removing internal ticket references as this is OSS repo

pkg/pipeline/source/sdk/appwriter_test.go
```diff
@@ -0,0 +1,243 @@
...
+// This is the state a track is in during the race. AddSourceBin links the new
+// bin and calls SyncStateWithParent (gstreamer/bin.go), so the element is on its
+// way to PLAYING via PAUSED while our Go code is still waiting for a bus message
+// that CloseWriters will cause to be dropped. Sending EOS from there works.
+//
+// Boundary worth knowing: an appsrc left in NULL - never asked to change state -
+// accepts EndStream() with FlowOK but never propagates it, and the pipeline hangs.
+// egress never links a bin without syncing its state; if that ever changes, EOS
+// alone will not be enough to unblock shutdown.
+func TestEOSFromNotYetPlayingAppsrc(t *testing.T) {
```
milos-lk: Beside demonstration/education purposes  - I am not sure what's the value of these tests  and maintaining them long term - as far as I can see no prod egress code is under test here - we can remove all production changes made here and they would probably still be passing - right?
Also testing pipeline appsrc has different values for some properties (e.g is-live) which makes it even further from behaving the same way prod pipeline will.

milos-lk:
The changes (with some comments on comments verbosity and tests) are generally good step forward. While reviewing the change I noticed a few more pre-existing races which are rather exotic - maybe worth following up separately.

pkg/pipeline/source/sdk/appwriter.go
pksgit: Done

pkg/pipeline/source/sdk/appwriter.go
pksgit: Done

pkg/pipeline/source/sdk/appwriter_test.go
pksgit: Done

pkg/pipeline/source/sdk/appwriter_test.go
pksgit: Agree. I have remove these tests.

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -665,9 +677,27 @@ func (w *AppWriter) maybeCheckPipelineLag(pts time.Duration) {
...
 
+// MarkAddedToPipeline signals that the appsrc is linked into the pipeline, which
+// blocks EOS aggregation on this writer until it ends its stream.
+func (w *AppWriter) MarkAddedToPipeline() {
+	w.addedToPipeline.Break()
+}
+
+// shouldSendEOS reports whether cleanup must push EOS into this writer's appsrc.
+//
+// The test is whether the appsrc is linked into the pipeline, not whether it was
```
milos-lk: nit: the function is small - so the comment is kind of stating the obvious.

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -665,9 +677,27 @@ func (w *AppWriter) maybeCheckPipelineLag(pts time.Duration) {
...
+// reported PLAYING: once CloseWriters() sets closing that notification is dropped,
```
milos-lk: nit: I would recommend against referencing caller sites here and having what I call comments coupling - implementation might change and this comment could quickly be outdated.

pkg/pipeline/source/track_worker_test.go
```diff
@@ -156,3 +156,76 @@ func TestStateTransitions_ActiveState(t *testing.T) {
...
+		select {
+		case op := <-w.opChan:
+			ops = append(ops, op.Type)
+		default:
+			return ops
+		}
+	}
+}
+
+func contains(ops []OpType, t OpType) bool {
```
milos-lk: nit: this function isn't needed. we have 2 options:
- built in slices contain
- testify's own require.Contains / require.NotContains

pkg/pipeline/source/sdk/appwriter.go
```diff
@@ -268,7 +270,17 @@ func (w *AppWriter) start() {
 	}
 
 	// clean up
-	if w.playing.IsBroken() {
+	if w.shouldSendEOS() {
+		if !w.playing.IsBroken() {
+			// Linked into the pipeline but never reported PLAYING: the CloseWriters
```
biglittlebigben: This also references calling site.

milos-lk (approved):

PR #1366, opened 2026-09-01
Add configurable timeout for web egress start signal

diazweb:
Web egress waits indefinitely for the page to log `START_RECORDING` before starting sinks. If that signal never arrives (e.g. the recorder page never logs it), the egress hangs forever with no way to self-recover — we hit this in production, stuck for days until manually stopped.

This adds an optional `session_limits.await_start_signal_timeout` config. When set, the egress aborts with `MsgStartNotReceived` if the start signal doesn't arrive in time, instead of waiting forever. Defaults to `0` (disabled), so existing behavior is unchanged unless configured.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ session_limits: # optional egress duration limits - once hit, egress will end wi
   file_output_max_duration: 1h
   stream_output_max_duration: 90m
   segment_output_max_duration: 3h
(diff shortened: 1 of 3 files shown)


diazweb:
@frostbyte73 how can we make CI pass?

PR #1378, opened 2026-09-08
Automate publish-chrome: weekly cron + no-op when current

frostbyte73:
publish-chrome.yaml runs unattended once a week: it resolves the current Chrome stable release, exits without provisioning anything if livekit/chrome-installer already has that tag, and otherwise builds and pushes it. Manual workflow_dispatch with an explicit version still works exactly as before

### DIFF
--- a/.github/workflows/publish-chrome.yaml
+++ b/.github/workflows/publish-chrome.yaml
@@ -1,11 +1,13 @@
 name: Publish Chrome
 
 on:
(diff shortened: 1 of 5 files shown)


biglittlebigben (approved):
Thanks!

As a follow up, we should use renovate to create a PR to update the chrome image version we use in the egress image (it may actually just work without any extra conf...)

PR #1382, opened 2026-09-09
Pin the GStreamer version once in .gst-version

biglittlebigben:
Same change as livekit/ingress#494, applied here. **The GStreamer version is unchanged at 1.24.12** — this is only about where the number lives.

It was previously spelled out in five places: the `gstVersion` const in `magefile.go` and four `FROM livekit/gstreamer:1.24.12-*` lines across `build/egress/Dockerfile` and `build/test/Dockerfile`. Nothing tied them together, so they could drift apart silently — that is exactly what had happened in ingress.

`.gst-version` is now the single source of truth, sitting alongside `.go-version`. Both Dockerfiles take the version as a `GSTVERSION` build arg with **no default** — a default would be a second copy and would drift, and without it the build fails at parse time instead of quietly using the wrong base image. The magefile and the two workflows that build these images (`publish-egress`, `test-integration`) read the file and pass it in.

`buildGstreamer` reads the same file, so the version this repo publishes and the version it builds against stay one number. Renovate can only ever offer versions that are already on Docker Hub, so publishing a brand-new GStreamer still starts with a deliberate edit to `.gst-version` followed by the `publish-gstreamer` workflow; the manager just closes the loop afterwards.

Renovate maintains the pin through a custom manager. `livekit/gstreamer` publishes no bare `X.Y.Z` tags — only `-dev`/`-prod`/`-base`/`-prod-rs` and arch variants — so a plain docker lookup would find nothing and fail silently; the manager tracks the `-dev` tags and `extractVersionTemplate` strips the suffix to keep the file a bare version. It is our own image, so it is exempt from the repo-wide two-week release quarantine, and a patch bump and a minor bump each get their own approvable PR.

## Heads-up

This repo is behind the published GStreamer tags, so once this merges Renovate will open two PRs — `1.24.13` (patch) and `1.26.11` (minor). Neither is applied here; take or close them as you see fit.

## Verification

- Ran Renovate against a scratch copy of this config at the real pin: it produces exactly those two updates, on `renovate/livekit-gstreamer-1.24.x` and `renovate/livekit-gstreamer-1.x`.
- `renovate-config-validator` passes.
- `docker buildx build --check` on both Dockerfiles with the arg resolves `livekit/gstreamer:1.24.12-dev`; without the arg both fail immediately at parse time.
- Built the `builder` stage of `build/test/Dockerfile` (the stage CI builds) against the resolved base.
- `actionlint` on both workflows: no findings on the new steps or their `steps.gst-version.outputs.version` references.
- `go vet -tags mage` passes; no hardcoded GStreamer version remains anywhere in the repo.

BuildKit emits a cosmetic `InvalidDefaultArgInFrom` lint warning per default-less `ARG` (two per Dockerfile, visible as build annotations). The builds succeed; a placeholder default would silence it but reintroduce a second copy of the version, so the Dockerfile comments explain why the default is deliberately absent.


boks1971 (approved):

.gst-version
```diff
@@ -0,0 +1 @@
+1.24.12
```
milos-lk, line 1: should we add it to the list of path for integration test workflow?

biglittlebigben: Yes — added, `.gst-version` is now in the Integration Test workflow's paths filter (42fe51f), so a GStreamer bump triggers the integration run.

milos-lk (approved):

PR #1393, opened 2026-09-17
Track action versions, exempt our own actions, group pion updates

biglittlebigben:
Two Renovate changes, both already landed in `cloud-ingress` (#279, #280).

### `helpers:pinGitHubActionDigestsToSemver`

Extends the preset we already used — `pinDigests: true` still applies, so every action
keeps running an immutable SHA. What it adds is an `extractVersion` + `versioning` pair
that tracks the full `vX.Y.Z` tag behind the digest instead of the mutable major:

```yaml
uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7      before
uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1  after
```

The trailing comment is what a reviewer reads when judging a workflow diff, and the exact
version is what gives an update PR a real changelog range and a `patch`/`minor` update
type instead of an opaque digest-to-digest bump.

### No quarantine for our own actions

`minimumReleaseAge` exists so a third-party release sits in the wild before we adopt it.
Our own actions have nothing to wait out — and for any referenced by branch the quarantine
does real harm, because Renovate ages a branch ref against the head commit of that branch.
`livekit/slack-notifier-action` can therefore only move in a fortnight where nobody pushed
to its default branch.

```json
{
  "matchManagers": ["github-actions"],
  "matchPackageNames": ["livekit/**"],
  "minimumReleaseAge": null
}
```

This mirrors the exemption `github.com/livekit/**` already has under `gomod`. The matcher
is bare `owner/repo`, not `github.com/livekit{/,}**` — the `github-actions` manager sets
`packageName` that way, so a module-style prefix would silently match nothing. It sits
after the `github workflows` group rule so its `minimumReleaseAge` wins, and leaves
`groupName` untouched so first-party actions still ride in the same group PR.

### `pion deps` group

This repo has 2 direct `github.com/pion/**` requires. The pion modules are co-released and
depend on each other, so bumping them one at a time produces PRs that don't build. Same
rule `protocol`, `cloud`, `livekit`, `server-sdk-go` and `cloud-connector` already carry.


paulwe (approved):

biglittlebigben:
> the change will prevent `@main`-pinned actions from updating

Right about the mechanism, and worth spelling out, but it doesn't apply to this PR — the two changes here cancel it.

`helpers:pinGitHubActionDigestsToSemver` sets its `versioning` on *every* `action`/`workflow` dep, not only the ones it can parse. For `livekit/slack-notifier-action@<sha> # main` that overrides the `exact` versioning the manager assigned, `isValid("main")` becomes false, and `lookupUpdates()` takes its "unsupported/unversioned value" path — which never fetches releases. The digest update is still created (`pinDigests` keeps the dep alive) but carries no `releaseTimestamp`. `minimumReleaseAgeBehaviour` defaults to `timestamp-required`, so a missing timestamp reads as "not old enough", and with the default `internalChecksFilter: strict` the update is marked `pendingChecks` on every run and never lands.

So the preset on its own would freeze branch-referenced actions at their current SHA. That is exactly what the second half of this PR prevents:

```json
{
  "matchManagers": ["github-actions"],
  "matchPackageNames": ["livekit/**"],
  "minimumReleaseAge": null
}
```

With no minimum in the merged config, `checkMinimumReleaseAge()` returns nothing pending and the digest update flows. The exemption is load-bearing here, not a convenience — the two changes have to ship together, which is why they're in one PR.

I checked the blast radius across all 116 action refs in the repos getting this change. Every ref that goes unversioned under the new versioning is livekit-owned, so every one is covered by the rule: `slack-notifier-action@main` (11 repos), `build-and-deploy-action@main` incl. its `/build` and `/deploy` subpaths (4 repos), `release-actions@main` (cloud-io). No third-party action is tracked by branch, and none carries a two-part `# v4.1` comment, which would hit the same path — the preset's regex accepts `v4` and `v4.1.2`, not `v4.1`.

The residual hazard is future-tense and worth knowing: add a third-party action referenced by branch, and it will stall silently unless it gets its own `minimumReleaseAge: null`. That's now written into our Renovate skill.

PR #1399, opened 2026-09-21
DO NOT MERGE: log raw sender report NTP/RTP pairs for av-sync diagnosis

milos-lk:
**Debugging only. Not intended to be merged.** Opened as a draft so it runs CI without looking like a candidate for review.

## Why

`TestEgress/File/1/TrackComposite/VP8` fails intermittently on `av-sync p80 exceeds 300ms tolerance`. The cause is not in this repo: the same commit measured 31.2ms on 09-15 and 333.2ms when rebuilt and re-run on 09-21.

The failing recordings show a constant audio offset, not a lossy path:

| | failing session (09-21) | passing session (09-15) |
|---|---|---|
| audio SR drift | 317.8ms, stable to 0.6ms over 35s | 818µs |
| video SR drift | within ±1.5ms | -721µs |
| event presence | 100% audio, 100% video | 100% |
| NTP estimator | 3 outliers rejected, 1 rebuild, audio never transitions to NTP | clean |

The six other audio tracks in that same job, same runner and same media nodes, all measured under 2ms. So the offset is specific to the session, sits on one track, and holds constant, which rules out network delay.

## What this adds

One log line per RTCP sender report, at the point egress already owns the callback (`pub.OnRTCP`), recording `trackID`, `kind`, `mime`, `ssrc`, `rtpTime`, `ntpTime`, `receivedAt`, `ntpMinusReceived`, `packetCount` and `octetCount`. Roughly 12 lines per track per test.

The synchronizer today exposes only the drift it derives from these reports (`OnSenderReport(func(drift time.Duration))`), which cannot separate the two remaining explanations.

## What the output decides

- `ntpMinusReceived` differing by ~318ms between the audio and video track of one participant, stable per track: the sender reports carry an offset NTP/RTP pairing, so the problem is upstream in SR generation.
- `ntpMinusReceived` consistent across both tracks but `ntpTime` vs `rtpTime` not fitting a line for audio: the pairs are noisy and `NtpEstimator` anchored badly, which points at the 3σ-over-100µs outlier gate in `ntpestimator.go`.

`go.mod` moves `github.com/pion/rtcp` from indirect to direct, which is what `go mod tidy` does for the new import.


milos-lk:
Pushed `859376f`: forces `enable_sync_engine: false` and `audio_tempo_controller.enabled: false` in code (`forceSyncDebugOverrides`, called from both `NewServiceConfig` and `NewPipelineConfig`), since the values come from the CI config secret and cannot be flipped from the repo. Both processes log `sync debug overrides applied` so the run proves the override took effect.

This is the A/B for the previous run's finding. The sender reports there were clean: audio and video arrived at the same instant with the same NTP timestamp, `ntpMinusReceived` agreeing to ~60µs, and each track's RTP advance matching its clock rate. The two tracks' `sessionPTS` also agreed to sub-millisecond at every report. What did not agree was the audio drift, a flat 731ms for the whole recording while video sat at ±1.7ms, because with the tempo controller enabled audio is emitted on wall-clock PTS while video transitions onto the NTP timeline.

With both settings off, SDK sources take the classic synchronizer path with audio PTS adjustment. If `TestEgress/File/1/TrackComposite/VP8` now passes consistently, the fault is in the sync engine's wall-clock audio path and the tempo controller failing to close the gap. If it still fails at ~330ms, that path is exonerated and the cause sits further upstream.

Still debugging only, still not for merge.

PR #1404, opened 2026-09-22
config: let a caller tee the log stream, and the handler's structured lines

paulwe:
A deployment that ships egress's logs somewhere has nowhere to install a tee: `InitLogger` builds the zap logger itself, and its variadic slot is already spoken for by the logger values. `LoggerTee` carries one in on the config instead. The zero value builds no core, so existing callers are unaffected.

The handler subprocess is the other half. Its own structured lines are passed straight through to the parent's stdout rather than its zap core, so a tee on the parent never sees them. `NewHandlerLoggerWithSink` hands them to a caller-supplied sink instead, which keeps the decision about where they go with the parent that owns the transport.

Requires protocol at or past `logger.WithTee`, hence the bump.

Consumed by livekit/cloud-egress, which uses both halves to publish its logs to the observability warehouse.


cnderrauber (approved):

boks1971 (approved):

paulwe:
Blocked on livekit/protocol#1813.

`go.mod` is pinned to that PR's head (`v1.52.1-0.20260922072557-d77b4a5b1c25`), a commit that only exists on `paul/logger-config-yaml`. Re-pin to protocol `main` once 1813 lands; do not merge this before then.

Background: the protocol bump in this PR made `logger.Config` the `ComponentLevelResolver`, which locks it on every resolve. `BaseConfig.Logging` is that same instance and `launchProcess` marshals it for the handler, so yaml boxing the struct read the mutex unsynchronized — a data race that failed the first subtest of all 8 integration jobs. 1813 fixes it upstream with a `yaml.Marshaler` that snapshots under the lock.

